PR #313, opened 2025-02-03
Remove "Music" Text From About page

paterasPOK:
Fix: Removed Music text from About page


mostafaalagamy:
thanks :)

AndreasBozoglou:
you can delete the issue here: https://github.com/mostafaalagamy/Metrolist/issues/276

PR #349, opened 2025-02-13
Updated strings.xml (Ukrainian translation)

zhabafrog:
The translation has been improved and completed (if I haven't missed anything) and is now up to date with the English version, I'll check it later. In addition, the order of the code was changed to match the English version (there were some inconsistencies which complicates translation).


zhabafrog:
found a small mistake, going to fix it right now

PR #383, opened 2025-03-08
Improve Lyric Auto-Scrolling Behavior

Aguga2201:
**This PR enhances the way lyrics auto-scroll to make it feel more intuitive and similar to Spotify.**

What Changed?
- **Smarter syncing behavior:** Instead of blindly jumping to the current line after 4 seconds of inactivity, auto-scroll now only re-engages when the lyrics are positioned within a specific sync zone between 1/2 and 7/8 of the screen.
- If the user scrolls away and the line isn't in the sync zone anymore, it won’t automatically pull them back.
- If the user scrolls to the lyrics but they appear in the top half, it will wait until the lyrics reach the middle before following along.
- If the lyrics are at the bottom, they will move back to the middle after a short delay.
- **Immediate Sync on New Line:** When the lyrics are synced to a new line (i.e. the current line is within the sync zone), the view now scrolls to that line immediately instead of waiting for the fallback delay. This ensures the view quickly follows the synced line when a new one is reached. The fallback delay (now reduced to 2 seconds) is still in place, but this change smooths the transition when syncing to new lines.

Why?
- Makes scrolling feel more natural and less jarring.
- Prevents unwanted jumps when manually navigating lyrics.

### DIFF
--- a/app/src/main/java/com/metrolist/music/ui/component/Lyrics.kt
+++ b/app/src/main/java/com/metrolist/music/ui/component/Lyrics.kt
@@ -23,6 +23,7 @@ import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


mostafaalagamy:
Thank you for fixing this and for your time as well.

PR #416, opened 2025-03-27
Update Italian translation

airon90:
* Updated full_description.txt with some fixes
* Updated strings.xml with new keys and fixed old ones
* Added short_description.txt


mostafaalagamy:
Thanks bro

Issue #434, opened 2025-03-29
Crash when creating synced playlist (when logged out)


Ao1Pointblank:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

1. Make sure YT login is disabled/logged out 
2. Search for any online playlist/album
3. Attempting to create a synced playlist from this content will crash the app without explanation.

### Expected behavior

I expected a warning that sync is only available when logged in, or perhaps that syncing would be available without an account and just work.

### Actual behavior

crash.
log states:
error 401
You must be signed in to perform this operation.

### Logs

type: crash
osVersion: google/redfin/redfin:14/UP1A.231105.001.B2/2025021000:user/release-keys
flags: dev options enabled
package: com.metrolist.music:113
process: com.metrolist.music
processUptime: 91363 + 621 ms
installer: com.looker.droidify

I4.g: Client request(POST https://music.youtube.com/youtubei/v1/playlist/create?prettyPrint=false) invalid: 401 . Text: "{
  "error": {
    "code": 401,
    "message": "You must be signed in to perform this operation.",
    "errors": [
      {
        "message": "You must be signed in to perform this operation.",
        "domain": "global",
        "reason": "unauthorized"
      }
    ],
    "status": "UNAUTHENTICATED"
  }
}
"
	at I4.l.r(SourceFile:232)
	at I4.l.j(SourceFile:13)
	at I4.D.b(SourceFile:108)
	at C4.c.r(SourceFile:737)
	at A5.a.l(SourceFile:9)
	at V5.J.run(SourceFile:110)
	at V5.U.c0(SourceFile:76)
	at V5.B.C(SourceFile:105)
	at a4.T.r(SourceFile:35)
	at A5.a.l(SourceFile:9)
	at V5.J.run(SourceFile:110)
	at x3.s.run(SourceFile:26)
	at d6.j.run(SourceFile:3)
	at d6.a.run(SourceFile:94)
	Suppressed: b6.e: [z0.u0@2329ed9, N.g@c9f09e, s0{Cancelling}@57fc77f, Dispatchers.IO]

### Metrolist version

11.2.0

### Android version

Android 14

### Additional information

If there is a limitation that prevents playlists from being synced without a login, that is fine.
[...]

PR #437, opened 2025-03-29
Fix: issue #434 syncing issue 

aalrehn:


mostafaalagamy:
Thanks for fixing this

PR #509, opened 2025-04-10
Update Italian translation

airon90:
Updated Italian strings.xml


airon90:
@mostafaalagamy I fixed some errors and added new translations. It should be ok now

mostafaalagamy:
Okay, i will merge it now, thanks!

PR #546, opened 2025-04-17
Add MediaButtonReceiver to support external media control

MetalDroid:
This pull request adds the MediaButtonReceiver declaration in the AndroidManifest.xml to allow the app to respond to external media button events (e.g., Bluetooth headsets, Tasker, car controls, etc.).

This does not modify any other part of the app's logic, just enables compatibility with system-level playback controls.

Tested and working with Bluetooth headset and Tasker.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -151,6 +151,14 @@
             </intent-filter>
         </service>
 
+        <receiver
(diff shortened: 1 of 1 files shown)


mostafaalagamy:
I don't know why the build is failing!, In any case, I’m going to merge it now. Thank you.

PR #623, opened 2025-04-30
feat: support editing both song title and artist name

mostafaalagamy:


zettaexa:
This is a great feature, but can you please make the artist name edit only effect the song rather than everything the channel has made?

mostafaalagamy:
> This is a great feature, but can you please make the artist name edit only effect the song rather than everything the channel has made?

done, try it now and tell me if there is anything else!

Note: build is merged, take backup before installation

zettaexa:
Ok there are a few issues ...
1. The edited artist name doesn't show up in the media player notification, as it only shows the original name

Metrolist where the artist name is edited
Screenshot_20250430-200135

On the media player itself
Screenshot_20250430-200149

This can lead to incorrect scrobbles for apps like pano scrobbler as it uses that notification to decide for artist name and song title

2. For songs that have 2 artist names attached, only the first one is changed rather than the whole string on the player

Such as this song
Screenshot_20250430-200906

On the edit menu it only shows the first artist mentioned in the track here
Screenshot_20250430-200953

And when you edit it to something else, it would appear correctly outside of the app
Screenshot_20250430-201101

But in the player it doesn't display correctly, and it makes a mess
Screenshot_20250430-201212

zettaexa:
Also one feature that would be good to add is to be able to select multiple songs to edit them, and for this feature there should be a (keep) tag / lock icon that keeps the original song name / artist name so it doesn't change every song that's being edited
This would be great for mass tagging songs as it would save tons of time

zettaexa:
Alr thanks for fixing all the issues i had with this feature!!! This is an amazing add to the app (if you add multi edit then it would be stellar but i think it's fully ready to ship for a stable release)

mostafaalagamy:
> multi edit

Multi edit needs comprehensive changes throughout the application, and this is difficult now because I am not working from a computer at the moment!

zettaexa:
> > multi edit
> 
> Multi edit needs comprehensive changes throughout the application, and this is difficult now because I am not working from a computer at the moment!

Okay makes sense! I don't mind it much without it

PR #666, opened 2025-05-10
Fix deep link handling when the app is fully closed

MetalDroid:
Fixes deep link handling to work properly even if the app is completely closed.


MetalDroid:
Deleted.

needs review, not working properly

PR #677, opened 2025-05-11
strings: separate upstream strings from metrolist

lucasmz-dev:
Allows for a Weblate to be properly set up, easily, and translations from upstream to be recycled.
Updates many translations


lucasmz-dev:
Basically https://github.com/Malopieds/InnerTune/pull/474 but for here
I notice that it seems Metrolist was forked from Malopieds' fork but Git history reset, if this is merged on both, we can potentially recycle translations between them as well. I also notice that some strings seem to be present in OuterTune. Same there. Though, I would need to probably only copy strings that are actually used, not anything else as it would be probably very spammy, and as I notice the string names used by upstream InnerTune and forks have this issue where they're too unspecific so similar things may collide. I hope to fix this upstream... though it could be a bit of an issue to also fix that in forks... but it'd not be too hard probably.

lucasmz-dev:
thanks for triggering CI, I'm depending on it to build and lint stuff :)

lucasmz-dev:
it seemed to complain about the xml line `<?xml version='1.0' encoding='utf-8'?> in the source file which is really weird to me.. but ok removed it...?

app/src/main/res/values/metrolist_strings.xml
```diff
@@ -0,0 +1,151 @@
...
+    <string name="numbers">Numbers</string>
+    <string name="subscribers">Subscribers</string>
+    <string name="views">Views</string>
+    <string name="likes">Likes</string>
+    <string name="dislikes">Dislikes</string>
+    <plurals name="seconds">
+        <item quantity="one">1 second</item>
+        <item quantity="other">%d seconds</item>
+    </plurals>
+</resources>
```
lucasmz-dev, line 151: The extracting process got rid of the comments in the metro-specific strings, I didn't re-organize here because I'm not really sure if just copying the comments would make sense, or if you might wanna re-organize them as you see fit.

Translations will always be kind of messy... Weblate for instance will always just follow like the order of translation of a specific string, so unless you automate it to like fix that, it will be messy, but that's mostly fine

lucasmz-dev:
it built! hot girl shit

lucasmz-dev:
something seems off, pt-BR isn't applying

lucasmz-dev:
oh, it seems I need to select the language in settings, but that setting seems quite broken... it preferred/only used? pt strings instead of pt-BR, it also doesn't seem to match the app language preferences settings... this feature seems quite broken... not really the target of this PR though... so keeping it as ready

lucasmz-dev:
preferably when merging do not use squash, there's some renaming and it would probably break that and instead seems to make it think some files are being deleted and just other ones added because they also change

Dark25:
could also be used i18n

lucasmz-dev:
people often confuse both and it's just kind of really stupid naming scheming to try and seem superior or some shit, not a fan
https://en.wikipedia.org/wiki/Internationalization_and_localization

Dark25:
> people often confuse both and it's just kind of really stupid naming scheming to try and seem superior or some shit, not a fan
> https://en.wikipedia.org/wiki/Internationalization_and_localization

Sorry, I didn't mean to offend, it was just a suggestion, but the implementation you've done to separate things seems very good to me.

lucasmz-dev:
> Sorry, I didn't mean to offend

You didn't DW and I didn't mean to for you too, but I mean the i18n terminology is very lost, [I've personally also not used it correctly in the past](https://github.com/mjaakko/NeoStumbler/commit/d68a71cbe42884dba9c2f5cfe3fbacf3927558c5). People don't know what is i18n or l10n anyway, I don't think I've messed with i18n either here... it would've meant to un-hardcode things, here I just organized the string files so that they can be managed easily

mostafaalagamy:
Ready to merge now?

lucasmz-dev:
Yeah nothing else AFAICT.

mostafaalagamy:
> Yeah nothing else AFAICT.

Thank you so much for your time.

lucasmz-dev:
@mostafaalagamy you are now able to set up a Weblate project (or whatever else you may want to use for a translation platform) for Metrolist. I could help but you'd need to be the one asking for the hosting as I seem to have reached the limit :) as I manage InnerTune as well I can share components such as the Glossary or others into Metrolist (provided I'm an admin or have whatever relevant permissions are for that...?)

PR #716, opened 2025-05-26
Translation updates from Codeberg Translate

lucasmz-dev:


lucasmz-dev:
these 'added' commits are because of the 'add missing languages' which makes languages match, it gives a more realistic translation percentage, but we can undo that or simply squash

PR #744, opened 2025-05-29
feat: add filter chips

FineFindus:
Adds the filter chip list shown at the top of the homepage, allowing users to filter the homepage by mood/genre.

Port of https://github.com/OuterTune/OuterTune/pull/508.

Depends on https://github.com/mostafaalagamy/Metrolist/pull/743. It's marked as a draft, as I'm not quite sure on how to best integrate it in the UI, considering there is already a chip row.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/HomeScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/HomeScreen.kt
@@ -43,6 +43,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
@@ -175,6 +176,18 @@ fun HomeScreen(
             backStackEntry?.savedStateHandle?.set("scrollToTop", false)
         }
     }
+
+    LaunchedEffect(Unit) {
+        snapshotFlow { lazylistState.layoutInfo.visibleItemsInfo.lastOrNull()?.index }
+            .collect { lastVisibleIndex ->
+                val len = lazylistState.layoutInfo.totalItemsCount
+                if (lastVisibleIndex != null && lastVisibleIndex >= len - 3) {
+                    viewModel.loadMoreYouTubeItems(homePage?.originalPage?.continuation)
+                }
+            }
+    }
+
+
 
     val localGridItem: @Composable (LocalItem) -> Unit = {
         when (it) {
@@ -735,7 +748,7 @@ fun HomeScreen(
                     }
                 }
             }
-            if (isLoading) {
+            if (isLoading || homePage?.originalPage?.continuation != null) {
                 item {
                     ShimmerHost(
                         modifier = Modifier.animateItem()
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/HomeScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/HomeScreen.kt
@@ -1,6 +1,7 @@
 package com.metrolist.music.ui.screens
 
 import android.annotation.SuppressLint
+import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.combinedClickable
 import androidx.compose.foundation.gestures.snapping.rememberSnapFlingBehavior
@@ -148,6 +149,7 @@ fun HomeScreen(
 
     val allLocalItems by viewModel.allLocalItems.collectAsState()
     val allYtItems by viewModel.allYtItems.collectAsState()
+    val selectedChip by viewModel.selectedChip.collectAsState()
 
     val isLoading: Boolean by viewModel.isLoading.collectAsState()
     val isRefreshing by viewModel.isRefreshing.collectAsState()
@@ -187,7 +189,12 @@ fun HomeScreen(
             }
     }
 
-
+    if (selectedChip != null) {
+        BackHandler {
+            // if a chip is selected, go back to the normal homepage first
+            viewModel.toggleChip(selectedChip)
+        }
+    }
 
     val localGridItem: @Composable (LocalItem) -> Unit = {
         when (it) {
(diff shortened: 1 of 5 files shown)


mostafaalagamy:
I've been trying to come up with an idea for creating filters for several days, and I'm still searching! If you have any ideas, you can let me know.

FineFindus:
Personally, I would remove the current bar and replace it with the chip row in this PR
- 'Liked', 'Downloaded', 'Cached' and 'Account' are already available in the Library tab.
- History and Stats could be moved to an overflow menu.

This would improve the UX as the current chip row acts more like a collection of buttons. This breaks user expectations of how UI elements ought to work, based on experience with other apps and other screens in this application which use them as filters.

mostafaalagamy:
> Personally, I would remove the current bar and replace it with the chip row in this PR

Okay, can move history and stats to library?

zettaexa:
Can this be optional? I like the chips on top of the home page, it's consistent with most InnerTune apps

FineFindus:
> Okay, can move history and stats to library?

Tested in https://github.com/mostafaalagamy/Metrolist/pull/744/commits/39a8c846ea2b1a1d65c697d7683ea2711ae513dc, however I don't think that that's a great solution, as both History and Stats are not playlist/library items, but much more 'management' items. Also, with the way I've added them, we're essentially back to the way of badly replicating buttons in chips.

mostafaalagamy:
> Tested in 39a8c84

I noticed that it’s bad after testing now. It’s possible to mimic YouTube Music in this and have them in the top app bar, or the other solution is to place them before the mood filters so the order would be: history, stats, other chips filters. What do you think of this?

zettaexa:
> > Tested in [39a8c84](https://github.com/mostafaalagamy/Metrolist/commit/39a8c846ea2b1a1d65c697d7683ea2711ae513dc)
> 
> I noticed that it’s bad after testing now. It’s possible to mimic YouTube Music in this and have them in the top app bar, or the other solution is to place them before the mood filters so the order would be: history, stats, other chips filters. What do you think of this?

Great idea!! maybe make a separator between the two types of chips and it would be good, Solves all issues

FineFindus:
Hmm, while I think that it is a viable alternative, I'm not super happy with it, as it would be still having the issues of using chips as both filters and buttons. Maybe we could style them as buttons?

Also, my YouTube app only displays history in Library  (as a menu item) and in the account menu.

mostafaalagamy:
> Hmm, while I think that it is a viable alternative, I'm not super happy with it, as it would be still having the issues of using chips as both filters and buttons. Maybe we could style them as buttons?
> 
> Also, my YouTube app only displays history in Library  (as a menu item) and in the account menu.

I am not satisfied with it either because the history and stats work differently than the filters. I am trying to restructure the homepage—I will inform you of any idea I come up with!

mostafaalagamy:
What do you think about using the top app bar for this?
We could make it so that clicking on the name of the currently open page (whether it's Home, Explore, or Library) triggers a dropdown menu containing history, Stats, and Account. A downward arrow icon next to the page name would indicate to the user that there is something interactive there.

mostafaalagamy:
> What do you think about using the top app bar for this?
> We could make it so that clicking on the name of the currently open page (whether it's Home, Explore, or Library) triggers a dropdown menu containing history, Stats, and Account. A downward arrow icon next to the page name would indicate to the user that there is something interactive there.  

Alternatively, we could add tabs for History, Stats, and Account below the top app bar, and place the filters beneath them. I believe this option might be the best solution!

mostafaalagamy:
> > What do you think about using the top app bar for this?
> > We could make it so that clicking on the name of the currently open page (whether it's Home, Explore, or Library) triggers a dropdown menu containing history, Stats, and Account. A downward arrow icon next to the page name would indicate to the user that there is something interactive there.  
> 
> Alternatively, we could add tabs for History, Stats, and Account below the top app bar, and place the filters beneath them. I believe this option might be the best solution!

like this:

zettaexa:
> What do you think about using the top app bar for this? We could make it so that clicking on the name of the currently open page (whether it's Home, Explore, or Library) triggers a dropdown menu containing history, Stats, and Account. A downward arrow icon next to the page name would indicate to the user that there is something interactive there.

Hmm that might be good but I'd prefer having both the history stats liked chips alongside the mood chips ...

mostafaalagamy:
>  having both the history stats liked chips

This idea was my own innovation, but it’s not official—the way it functions is completely different from how chips work

FineFindus:
> Alternatively, we could add tabs for History, Stats, and Account below the top app bar, and place the filters beneath them. I believe this option might be the best solution!

I think dropdown menu would be better as tabs would implicate that each of the options equally important. Also feel free to implement that already, I want really be able to work on this for the next couple of days :)

mostafaalagamy:
> > Alternatively, we could add tabs for History, Stats, and Account below the top app bar, and place the filters beneath them. I believe this option might be the best solution!
> 
> I think dropdown menu would be better as tabs would implicate that each of the options equally important. Also feel free to implement that already, I want really be able to work on this for the next couple of days :)

I have implemented the drop menu now!

zettaexa:
Can you also add links to the liked, downloaded, and cached playlists as well and give icons to the menu?

mostafaalagamy:
> Can you also add links to the liked, downloaded, and cached playlists as well and give icons to the menu?

I don't want it to be stupidly long, i can add icons for the items okay!

zettaexa:
... well that's what items where in the chips bar originally so

zettaexa:
maybe adding a category in the library for pages like this might be better than a menu but km not sure

mostafaalagamy:
> maybe adding a category in the library for pages like this might be better than a menu but km not sure

I don't think so, this is her natural place!

mostafaalagamy:
I have also added the mood filters. You can now try them in the latest commit from the 'filters' branch!

zettaexa:
Doesn't work when j click on them

mostafaalagamy:
> Doesn't work when j click on them

I'm just testing out the new design to see if we can improve it or not, I think it's stable now!

FineFindus:
> I'm just testing out the new design to see if we can improve it or not, I think it's stable now!

There seems to be an UI issue, where the appbar overlaps a fix pixels of the content, otherwise looks good.

mostafaalagamy:
Hey, as for moving history and stats to the library, we don't need this because we use the chips incorrectly when using them together.

mostafaalagamy:
> where the appbar overlaps a fix pixels of the content

I didn't catch it, can you explain this?

FineFindus:
> I didn't catch it, can you explain this?

Never mind I can't reproduce it anymore

mostafaalagamy:
> > I didn't catch it, can you explain this?
> 
> Never mind I can't reproduce it anymore

Thanks for your effort!

FineFindus:
You're the maintainer, so It's your decision :) But from my side yes.

mostafaalagamy:
> You're the maintainer, so It's your decision :) But from my side yes.

done, thanks!

PR #869, opened 2025-06-28
Improve Auto Scroll Lyrics Setting

Aguga2201:
Currently, turning off the "Auto Scroll Lyrics" setting doesn't fully disable lyric scrolling. It still scrolls to the current lyrics when they are on screen and stop only if the user manually scrolls away (the default behavior before the setting was implemented). When turned on, it would always scroll to the current lyrics. 

With this update, disabling the setting now completely stops all automatic lyric scrolling. When enabled, lyrics will scroll automatically only when the current line is visible, and will pause if the user scrolls manually.

I also fixed a little bug where it scrolled even when it shouldn't by adding a new check


mostafaalagamy:
Thanks for fixing this!

PR #895, opened 2025-07-03
Refactor: Only romanize lyrics when is Japanese and not just Chinese

idlukakas:
This commit refactors the lyrics processing logic to better distinguish between Japanese and Chinese text.

Previously, Japanese text detection could sometimes misidentify Chinese text as Japanese due to shared Kanji characters.

This change introduces an `isChinese()` function in `LyricsUtils.kt` that uses a heuristic based on the proportion of CJK Unified Ideographs to Hiragana/Katakana characters.

The `Lyrics.kt` component now uses this new `isChinese()` check in conjunction with `isJapanese()` to more accurately determine when to romanize Japanese lyrics, preventing incorrect romanization of Chinese text.


mostafaalagamy:
Great, thanks for your effort, man!

Issue #898, opened 2025-07-03
Chips Row in OnlineSearchResult is slightly cropped from the top


FullerBread2032:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

Go to `Search`
Search for any online song

### Expected behavior

Image shows ChipsRow with padding `AppBarHeight + 4.dp`
https://github.com/mostafaalagamy/Metrolist/blob/e098718f24ce1ff7cd7c0b299e9ec19740d9d45f/app/src/main/kotlin/com/metrolist/music/ui/screens/search/OnlineSearchResult.kt#L278

![Image](https://github.com/user-attachments/assets/451951bc-d4c0-4a58-8699-8a50cd0a179b)

### Actual behavior

This image shows actual code
ChipsRow seems to be clipped by search bar
![Image](https://github.com/user-attachments/assets/e7c107d9-22c4-477f-8fa3-1fa20c54d55e)

### Screenshots/Screen recordings

The fragment that was cut off by the search bar is marked in red.
![Image](https://github.com/user-attachments/assets/fe0e9569-6fd6-49a1-b4e3-bc56e8eaa2ba)

### Logs

Not applicable

### Metrolist version

12.1.0

### Android version

Android 13

mostafaalagamy:
solved now?

PR #906, opened 2025-07-04
fix(layout): prevent ChipsRow overlap, better approach

FullerBread2032:
Better solution for #898 bug

I don't know why the SearchBar height value was set to 100.dp
The correct value should be 88.dp

**24.dp** - Android status bar height + **64.dp** - App Bar height


mostafaalagamy:
> I don't know why the SearchBar height value was set to 100.dp\nThe correct value should be 88.dp

I don't remember why I set it to 100, but there was a purpose for this. However, if it has solved the problem, then that's fine!

PR #947, opened 2025-07-10
fix(MusicService): restore Queue on application restart

FineFindus:
Fixes an issue, whereupon application restart, the queue would fail to be restored. This is likely due to the player not yet being correctly initialized, and thus the queue file is deleted.

Closes: https://github.com/mostafaalagamy/Metrolist/issues/775

Marked as a draft, as it still requires more testing.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -1114,11 +1114,6 @@ class MusicService :
     }
 
     private fun saveQueueToDisk() {
-        if (player.playbackState == STATE_IDLE) {
-            filesDir.resolve(PERSISTENT_AUTOMIX_FILE).delete()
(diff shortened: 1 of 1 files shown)


mostafaalagamy:
- **Before the PR (fix)**: When removing the app from *Recent Apps* and reopening it, the playback queue and player would restore correctly.  
- **After the PR (fix)**: When removing the app and restarting it, the queue and player are no longer restored.

mostafaalagamy:
Hi, you can test the latest beta version now and let me know if it works normally!

FineFindus:
Sorry it took me so long to get back to this. No, unfortunately the that fix does not work for me.

mostafaalagamy:
> Sorry it took me so long to get back to this. No, unfortunately the that fix does not work for me.

Yes, it didn't work for me either when I closed the app several times. Now it works very well. Try latest beta version.

PR #1101, opened 2025-07-28
fix: modify shadow to respect the rounding

MicheDePainn:
I know it's just a little shadow.
Screenshot_2025-07-28-14-01-30-672_com metrolist music-edit


mostafaalagamy:
Good catch, thank you for your contribution and your time!

PR #1205, opened 2025-08-08
feat: Add ability for seek skip duration to be added up progressively

innixunix:
Needed this myself because i find the static 5s limit to be low 

Adds a new option on player settings which when enabled, adds 5 seconds to each seek skip incrementally on each tap until the clicking stops found this myself when listening to long songs but it could be useful for other use-cases too


mostafaalagamy:
Thank you for your contribution. It's a good idea, and I just tried it now—I think it's indeed useful for long songs. Great work, thank you and your time!

PR #1228, opened 2025-08-10
Audio Normalization: Amplify low-volume tracks

### LINKED ISSUE #1146: Better volume normalization
### Checklist

### Feature description

Currently, the volume normalizer reduces the volume of tracks that have a lot of loudness, but in the case of tracks with low volume, it does not increase the volume. 

The adjustment should do both, increase the volume or reduce it as appropriate.

To play all tracks at the same volume

### Additional information

_No response_

MetalDroid:
When normalizing audio, if the track has a low volume, apply a maximum gain of +3dB.

I think it's a fairly acceptable value; increasing it too much could be counterproductive.

Perhaps it should be accompanied by a switch for those who don't want to amplify?

Please review. There may be better ways to do this...

Closes #1146

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -229,6 +229,8 @@ class MusicService :
 
     private var consecutivePlaybackErr = 0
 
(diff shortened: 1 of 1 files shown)


mostafaalagamy:
Thank you for your contribution and your time. I will merge it now, and if someone requests to add an option in the future, I will add it!

PR #1263, opened 2025-08-12
tweak: only darken and show icon on Miniplayer if paused

ThatOneCalculator:
ref: https://github.com/mostafaalagamy/Metrolist/issues/1252#issuecomment-3177784291


mostafaalagamy:
Thank you for your contribution. I understand your perspective on this, and the design will be better. However, after implementing it, we will find that many people will request or tell me that the pause button should appear in the mini-player!  

Also, ordinary users or beginners won’t realize that there is a pause option in this location. What do you think?

ThatOneCalculator:
I feel like the new miniplayer already has two more implicit actions (swiping left/right to skip tracks), and having the album art be play/pause is intuitive enough – if at any point a new user pauses their music, they should immediately see the icon overlay on the album art regardless, making it very clear it's a button. If people complain, this should be reverted, but for now I think it's worth a chance to merge, I don't think new users will complain.

mostafaalagamy:
> I feel like the new miniplayer already has two more implicit actions (swiping left/right to skip tracks), and having the album art be play/pause is intuitive enough – if at any point a new user pauses their music, they should immediately see the icon overlay on the album art regardless, making it very clear it's a button. If people complain, this should be reverted, but for now I think it's worth a chance to merge, I don't think new users will complain. 

Ok, I'll merge it now! Thanks.

Issue #1260, opened 2025-08-12
The app crashes when I open the appearance settings.


NamekujiLSDs:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

1.Open app
2.Go settings
3.tap appearance setting

### Expected behavior

Open appearance settings

### Actual behavior

Crash app

### Logs

[metrolist.log](https://github.com/user-attachments/files/21725773/metrolist.log)

### Metrolist version

12.3.0

### Android version

Android 13(Lineage OS 20 bgN but with gapps)

FullerBread2032:
Hello!
I can reproduce this bug when I set app language to Japanese 

Fix: #1265 

You can check if you want if the issue still occurs in the debug version
https://github.com/mostafaalagamy/Metrolist/actions/runs/16907006789/artifacts/3743927116

PR #1265, opened 2025-08-12
fix: App crash when opening Appearance settings

FullerBread2032:
Fixes: #1260 
The string has been changed to be identical to the other language versions
The crash occurred when the app language was set to Japanese.


mostafaalagamy:
Thanks, i will merge it now!

Issue #1270, opened 2025-08-12
Albums stats


Showtc:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

1. Go to statistics 
2. Watch albums

### Expected behavior

We should see how many time we listened to an album with the total time

### Actual behavior

Everything is at 0

### Screenshots/Screen recordings

<img width="1080" height="857" alt="Image" src="https://github.com/user-attachments/assets/c304d1a0-fef9-4df6-9675-2a6f71372162" />

### Logs

No log

### Metrolist version

12.3.0

### Android version

Android 16 QPR1

PR #1271, opened 2025-08-12
fix: Album stats

FullerBread2032:
Fixes: #1270 

The bug appeared after commit [`c8cb17a`](https://github.com/mostafaalagamy/Metrolist/commit/c8cb17a51cffba15a8982ab1a32da436a7cf51c0), so I reverted to the state before it and removed the warnings during build


mostafaalagamy:
I have been working on it currently, thank you for doing this.

PR #1274, opened 2025-08-12
feat/refactor: option to combine pitch and tempo into one & save values on app exit

ghost:
Reworked the code responsible for the pitch and tempo changes

Now there's a new toggle in the popup called "Link pitch to tempo" which when enabled removes the pitch slider and uses the tempo value for both 

By default, this is disabled to match the previous behavior of having separate values for both

Also made both pitch and tempo be saved even on app exit

I added this because I prefer having one slider for both  

Closes https://github.com/mostafaalagamy/Metrolist/issues/370


cirnoe:
bump, also this should be the default behavior, provided the app is aimed towards music, and not really movies/stories

PR #1338, opened 2025-08-18
Widget Preview

Gian-Fr:
this is an INITIAL implementation of widget, aimed to get a feedback from community, it DOES NOT provide a working widget as of now.
It is based on the YT music widget
immagine

- [ ] Working Media control (play - pause - previous - next - like - dislike)
- [ ] Widget color change based on current song
- [ ] Show cover in widget

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -17,8 +17,8 @@ android {
         applicationId = "com.metrolist.music"
         minSdk = 26
         targetSdk = 36
-        versionCode = 124
-        versionName = "12.3.0"
+        versionCode = 125
+        versionName = "12.4.0"
 
         multiDexEnabled = true
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
@@ -124,6 +124,9 @@ android {
 
     lint {
         lintConfig = file("lint.xml")
+        warningsAsErrors = false
+        abortOnError = false
+        checkDependencies = false
     }
 
     androidResources {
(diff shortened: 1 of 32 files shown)


mostafaalagamy:
I think it looks very good! However, I believe we should replace the 'Like' button with a 'Favorite' button, and since there’s no 'Dislike' button in the app, it would be better to remove it entirely. I also think displaying the song’s cover art would be an good, but the feature that changes colors based on the song’s artwork isn’t really necessary. Let’s wait for everyone else’s feedback!

itsmeteddyhere:
I think you should have the ability to choose between the widget following dynamic color from wallpaper or following the album cover color. Other than that, it looks great!

ThatOneUnoriginal:
Been praying for the day that I'd see something like this. I have a suggestion. Since the dislike button would not be there — given there's no dislike feature in Metrolist — perhaps it could be replaced with a repeat button? As for the color of the widget, the default should be based on the wallpaper, though perhaps an option to have it based on the album cover (solid or perhaps also gradient) could be implemented as well.

Gian-Fr:
I could simply use material theme. I'll publish the new preview as soon as it is ready 😉

FullerBread2032:
Hi, I'd like to ask how the work on this PR is going, do you need any help?

Gian-Fr:
Hi! Thanks for asking. Actually I'm studying for an exam so the work is stopped as of now, I'll resume after 19 September.
I haven't done anything more than the actual draft

Majiild:
Widgets is what this app needs the most i think

Mono1912:
love the widget, one tiny request; please change the wording of the queue manager. instead of "song REMOVED FROM PLAYLIST" _please_ change it to "song removed from **queue** ". it would prevent MANY heart attacks.

Gian-Fr:
From now on I'll start working again on this!

mostafaalagamy:
> From now on I'll start working again on this!

Yes, please, because the branch was deleted due to some issues!

PR #1340, opened 2025-08-18
Update libs.versions.toml

Gian-Fr:
this is the latest version supported by Android Studio, needed to work with it


kkoshin:
The Compose team have started publishing alpha version of the [offical BOM](https://developer.android.com/develop/ui/compose/bom).

Could you also update the dependency method accordingly?

PR #1346, opened 2025-08-18
feat: Added app language support for versions below Android 13

### LINKED ISSUE #1154: Language changer doesn't show up
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Language changer doesn't show up

it's follow the system language
 ‎ 

Image

### Metrolist version

12.2.0

### Android version

Android 11

FullerBread2032:
Added support for changing app language for version below Android 13
In versions A13 and later, changing the application language is available in the settings

Closes #1154

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/settings/ContentSettings.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/settings/ContentSettings.kt
@@ -56,7 +56,6 @@ fun ContentSettings(
 
     val (contentLanguage, onContentLanguageChange) = rememberPreference(key = ContentLanguageKey, defaultValue = "system")
     val (contentCountry, onContentCountryChange) = rememberPreference(key = ContentCountryKey, defaultValue = "system")
-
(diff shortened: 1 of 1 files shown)


mostafaalagamy:
The issue is that the application's default language is English even when the device language is set to another language. However, if you change the language and then switch back to the system language, it works normally.

FullerBread2032:
I think the PR is ready. I fixed the bug you mentioned and also tweaked the code a bit to make it more readable.

However, I don't know how to fix the problem that after changing the language, not all strings change to the new language in ContentSettings.kt, but after exiting that screen, the application language is completely changed and works as expected.

mostafaalagamy:
> However, I don't know how to fix the problem that after changing the language, not all strings change to the new language in ContentSettings.kt, but after exiting that screen, the application language is completely changed and works as expected

This is normal!

Issue #1318, opened 2025-08-17
Custom order changes after restarting the phone.


Elekrys:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

1.Go to Playlist
2.Click on custom order
3.Click the lock icon to unlock playlist
4.Change the order of the songs in the playlet
5.Click the lock icon to lock playlist
6.Then go back and restart the phone

### Expected behavior

The playlist should remain the way it was changed after restarting the phone

### Actual behavior

It changes back to the way it was(Date added)

### Logs

N/A

### Metrolist version

12.4.0

### Android version

Android 15

FullerBread2032:
@Elekrys 
Hi,
Could you test this version?
I've added support for order synchronization. I'm curious if there are any bugs I haven't noticed. 
Thanks 🙌

https://github.com/mostafaalagamy/Metrolist/actions/runs/17067386235

Issue #1231, opened 2025-08-10
Song reordering doesn't work on synced playlists


Bindrr:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

1. Reorder songs in a playlist
2. Click on refresh button

### Expected behavior

Songs will be reordered on Youtube Music

### Actual behavior

Nothing happens, the song don't reorder at all

### Screenshots/Screen recordings

https://github.com/user-attachments/assets/af3ddee0-37fd-4aa8-b07a-a25c9601e55c

### Logs

no logs

### Metrolist version

12.3.0 (latest, but it has not been working on previous versions either)

### Android version

latest

ganairm:
can confirm, same thing happens to me

ThatOneUnoriginal:
I am assuming how the app communicates with YouTube / YouTube Music about playlist changes. I have noticed similar issues with removing songs from a playlist. It just isn't possible. A refresh of the sync will make the removed songs reappear as if they never left. So I'm guessing that's whats happening here. The playlist on YouTube Music isn't updated (for whatever reason) to reflect the positioning changes.

ganairm:
> I am assuming how the app communicates with YouTube / YouTube Music about playlist changes. I have noticed similar issues with removing songs from a playlist. It just isn't possible. A refresh of the sync will make the removed songs reappear as if they never left. So I'm guessing that's whats happening here. The playlist on YouTube Music isn't updated (for whatever reason) to reflect the positioning changes.

no you have to actually wait inside the playlist before exiting. the time it takes for each song to be updated is around 2-4 seconds so just wait a bit snd it will all be updated

Bindrr:
> I am assuming how the app communicates with YouTube / YouTube Music about playlist changes. I have noticed similar issues with removing songs from a playlist. It just isn't possible. A refresh of the sync will make the removed songs reappear as if they never left. So I'm guessing that's whats happening here. The playlist on YouTube Music isn't updated (for whatever reason) to reflect the positioning changes.

When changing the order of the songs, it doesn't matter whether you wait or not, and it doesn't matter when you press the sync playlist button (or if you don't press it at all). It doesn't work either way

ganairm:
[...]

PR #1354, opened 2025-08-19
feat: Sync playlist order with YouTube Music

FullerBread2032:
Added playlist custom order synchronization between Metrolist and YouTube Music.
Now if a user moves a song, the changes will automatically be reflected in the YT Music playlist

Closes: #1318 #1231

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/playlist/LocalPlaylistScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/playlist/LocalPlaylistScreen.kt
@@ -405,6 +405,25 @@ fun LocalPlaylistScreen(
                 database.transaction {
                     move(viewModel.playlistId, from, to)
                 }
+
+                // Sync order with YT Music
+                if (viewModel.playlist.value?.playlist?.isLocal == false) {
+                    viewModel.viewModelScope.launch(Dispatchers.IO) {
+                        val playlistSongMap = database.playlistSongMaps(viewModel.playlistId, 0)
+
+                        val successorIndex = if (from > to) to else to + 1
+                        val successorSetVideoId = if (successorIndex <= viewModel.playlistSongs.value.size - 1) playlistSongMap[successorIndex].setVideoId else null
+
+                        playlistSongMap[from].setVideoId?.let { setVideoId ->
+                            YouTube.moveSongPlaylist(
+                                viewModel.playlist.value?.playlist?.browseId!!,
+                                setVideoId,
+                                successorSetVideoId,
+                                )
+                        }
+                    }
+                }
+
                 dragInfo = null
             }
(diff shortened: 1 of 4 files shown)


FullerBread2032:
I think the PR is ready. 
But I'm not sure what to do if the changes are applied offline, then the playlist will be overwritten by what's on YT Music. 
What do you think?

mostafaalagamy:
I have tested the reordering on local synchronized playlists, and it works well. However, in the OnlinePlaylist fetched from YouTube Music, it does not work. I think there is no way to solve this problem because I tried to solve it a while ago and did not succeed! If it cannot be solved, the drag button should be hidden from the online playlist screens.

FullerBread2032:
Okay, could you tell me how I could reproduce this because I don't know where it's not working? 😅
1. I create a playlist on YTM
2. I add at least two items to it
3. I rearrange the items in Metrolist (from Library)
4. The changes are reflected in YTM

I am aware of the existence of a bug/missing playlist editing function when it is launched from the main menu in e.g. the `Your YouTube playlists` section, even though it belongs to the user

Also Lock IconButton only appear in `LocalPlaylistScreen.kt`, i don't see any song reordering function in `OnlinePlaylistScreen.kt`

mostafaalagamy:
Existing editable playlists do not have the feature working, while newly created synchronized editable playlists work perfectly!

The first playlist I opened was existing and synced from a YouTube account, the second playlist is the one I created new!

https://github.com/user-attachments/assets/5178d7c0-43af-4892-b252-a93def33bfa9

FullerBread2032:
Okay, tomorrow I'll try to recreate it and fix it if possible.

FullerBread2032:
Okay, tomorrow I'll try to recreate it and fix it if possible.

mostafaalagamy:
> Okay, tomorrow I'll try to recreate it and fix it if possible. 

Take your time!

FullerBread2032:
@mostafaalagamy 
I think it's timing. If we sync very quickly after the change, the playlist will be there before the POST request is sent, but if I press sync again, the playlist displays correctly.

mostafaalagamy:
> @mostafaalagamy 
> I think it's timing. If we sync very quickly after the change, the playlist will be there before the POST request is sent, but if I press sync again, the playlist displays correctly.

Yes, that's correct! But only in newly created playlists, not the existing ones

FullerBread2032:
Okay, I was wondering why this playlist was missing two icons for deleting and editing the playlist. 😅

I managed to reproduce this in a collaborative playlist. 
Metrolist shows that songs can be moved despite the lack of permissions (403 error)
I think drag handle should be only accessible when playlist is editable

mostafaalagamy:
> I think drag handle should be only accessible when playlist is editable

Exactly

Issue #1332, opened 2025-08-18
Highlighted lyrics in the centre


bibliophilical:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

The highlighted lyrics now sit at the centre of the device screen height but not the centre of the lyric window. This makes it looks visually located slightly below the expected place where user read lyrics at.

Vertically centre the highlighted currently played lyrics to the lyric window.

### Why do you want this feature?

Make it easier to read.

### Additional information

Make it somewhere around the red arrow I think.

<img width="1080" height="2320" alt="Image" src="https://github.com/user-attachments/assets/3e5c6b5f-ac05-49f3-9543-7f744aee37fc" />

Sravan-k-177:
hello, i would love to work on this.!

primesonaldo:
+1 for this. Would much prefer the currently playing lyrics to be in the center.

PR #1357, opened 2025-08-19
Fix issue #1332

Sravan-k-177:
solves issue 1332
The code is changed for lyrics to be displayed at 1/3 of the screen, instead of the previously used 1/2 of the screen.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -89,12 +89,9 @@ android {
         debug {
             applicationIdSuffix = ".debug"
             isDebuggable = true
-            signingConfig = if (System.getenv("GITHUB_EVENT_NAME") == "pull_request") {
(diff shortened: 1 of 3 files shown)


app/schemas/com.metrolist.music.db.InternalDatabase/22.json
FullerBread2032, line 1: I don't think this file should be changed. It also has nothing to do with PR.

app/build.gradle.kts
FullerBread2032, line 1: I don't think this file should be changed. It also has nothing to do with PR.
I understand this is for testing purposes, but it shouldn't be included in this PR

Sravan-k-177: sorry, i have accidentally includes those file, now i made a pull request with just the required changes.

PR #1359, opened 2025-08-19
fix: improve active line position in lyrics screen

Sravan-k-177:
This Displays the lyrics at 1/3 of the screen so the lyrics are more readable.


mostafaalagamy:
Thank you for your contribution, and thank you for your time!

Now the situation is good, but what do you think about making the beginning of the words and the shimmer start from the top of the screen instead of the middle of the screen?

Issue #618, opened 2025-04-29
proxy authorization


gmankab:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

hello, can you please add support for proxy authosization with login and password

### Why do you want this feature?

because my proxy has login and password

mostafaalagamy:
I don't know how to do this, so I'll leave the issue open for PR.

byMali5k:
+1

PR #1374, opened 2025-08-20
feat: Add proxy auth support

FullerBread2032:
This pull request adds support for proxy authentication in the application, allowing users to configure proxy credentials (username and password) and use them for HTTP and SOCKS proxies throughout the app.

Tested on HTTP proxy with auth, and SOCKS5 proxy with auth
**Proxy Authentication Support**

* Added `ProxyUsernameKey` and `ProxyPasswordKey` preferences, and updated the app logic to handle proxy credentials when a proxy is enabled, including setting HTTP Basic authentication or a system authenticator depending on proxy type.
* Added a proxy configuration dialog in the settings screen, allowing users to input proxy type, URL, username, and password. Replaced the previous inline proxy settings with a single "Configure proxy" entry.

## TODO:
- [X] Fix/Change UI setting
- [X] Test if all endpoints uses proxy with auth

## Known Issues:
- [X] TextFields are updated on change, causing input problems

Closes: #618

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/App.kt
+++ b/app/src/main/kotlin/com/metrolist/music/App.kt
@@ -37,6 +37,10 @@ import timber.log.Timber
 import java.net.Proxy
 import java.util.*
 
+import java.net.Authenticator
+import okhttp3.Credentials
+import java.net.PasswordAuthentication
+
 @HiltAndroidApp
 class App : Application(), SingletonImageLoader.Factory {
     // Create a proper application scope that will be cancelled when the app is destroyed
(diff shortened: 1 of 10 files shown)


mostafaalagamy:
> * TextFields are updated on change, causing input problems

The problem is in your proxy configuration dialog where you're directly updating preferences on every keystroke:

OutlinedTextField(
    value = proxyUrl,
    onValueChange = onProxyUrlChange, // This updates preferences immediately
    label = { Text(stringResource(R.string.proxy_url)) },
    modifier = Modifier.fillMaxWidth()
)

Solution: Use local state variables for the dialog and only update preferences when the user confirms

FullerBread2032:
Authentication is hideable

PR #1448, opened 2025-09-01
Update LyricsScreen.kt

hariom982:
Updated the lyrics screen to keep the screen awake when the lyrics are displaying.


mostafaalagamy:
Thanks for your contribution and time, I've already added this in latest commit.

PR #1508, opened 2025-09-10
feat: Add custom thumbnail on local playlists

FullerBread2032:
Adds the option to change the cover art for local (non-synced) playlists.
For local playlists, the note about the need to connect a phone number to YouTube Music is no longer displayed in the dialog; it is saved as a URI link to the file.
Changed the logic so that instead of saving to the previous `playlistThumbnail`, it now uses `overrideThumbnail`.
This PR also adds a default cover art for an empty local playlist.

If possible, please test this version


mostafaalagamy:
> please test this version

works well!

PR #1525, opened 2025-09-11
feat: Add option to remove song from remote history

FullerBread2032:
This PR adds option to remove song from remote history used in YT Music, by sending feedback token

> [!IMPORTANT]
>  `historyRemoveToken` is only available from HistoryPage and is not saved in db

TODO:
- [x] Remove item from list


FullerBread2032:
Everything works for me!

PR #1560, opened 2025-09-14
Let the user decide whether to check for updates

sadorowo:
## Description of this PR
I've added a new preference that allows users to enable or disable automatic update checks.
This is particularly helpful for users who update the app via stores like F-Droid, where updates might not be available immediately, while the latest version is already available on GitHub.

As a quick workaround, instead of implementing multiple checks throughout the codebase where the updater logic is used, I made a small modification to `LaunchedEffect` in `MainActivity`. When the user disables update checks, the app version is reset to the one defined in `BuildConfig`. I believe this is a more efficient approach than adding conditional checks in every place where the updater logic is applied.

I've also added a new string key, `check_for_updates`, and provided translations in English ~~and Polish (as I’m a native Polish speaker).~~
~~If you're using a translation platform like Weblate or something similar and believe this should be handled there instead, feel free to leave a comment. I'll remove the translation if needed.~~

## How has this been tested
I...
1. Generated my own keystore file
2. Temporarily downgraded `VERSION_NAME` in `build.gradle.kts`
3. Built debug variant (`gradlew assembleDebug`)
4. Toggled the new switch in the About section off and on

edit: I've moved the strings to "metrolist_strings.xml" file. If these strings should be in general strings.xml file, I'll revert the last commit.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/NavigationBuilder.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/NavigationBuilder.kt
@@ -68,6 +68,7 @@ import com.metrolist.music.ui.screens.settings.PrivacySettings
 import com.metrolist.music.ui.screens.settings.RomanizationSettings
 import com.metrolist.music.ui.screens.settings.SettingsScreen
 import com.metrolist.music.ui.screens.settings.StorageSettings
+import com.metrolist.music.ui.screens.settings.UpdaterScreen
 import com.metrolist.music.ui.utils.ShowMediaInfo
 import com.metrolist.music.utils.rememberEnumPreference
 import com.metrolist.music.utils.rememberPreference
@@ -311,6 +312,9 @@ fun NavGraphBuilder.navigationBuilder(
     }
     composable("settings/discord/login") {
         DiscordLoginScreen(navController)
+    }
+    composable("settings/updater") {
+        UpdaterScreen(navController, scrollBehavior)
     }
     composable("settings/about") {
         AboutScreen(navController, scrollBehavior)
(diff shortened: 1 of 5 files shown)


mostafaalagamy:
Thank you for your contribution. I think we can place it on another screen instead of the About screen to keep the design clean without distortion, or create a new settings page named "Updater." What do you think?

I was also thinking of doing this and enabling notifications when an update is released!

sadorowo:
Yeah, sure, I've put it on the About page because I was unsure where to put it.

sadorowo:
Screenshot_20250914-204035_Metrolist Debug
Screenshot_20250914-204039_Metrolist Debug

So I decided to create an Updater screen. Let me know where I should place the string key - in strings.xml or metrolist_strings.xml? As I can see keys for names of all settings pages are in strings.xml, but maybe these keys were used in the original app

mostafaalagamy:
> So I decided to create an Updater screen. Let me know where I should place the string key - in strings.xml or metrolist_strings.xml? As I can see keys for names of all settings pages are in strings.xml, but maybe these keys were used in the original app

This is good, and we will also place enable notifications below it.
For any new feature or any addition, the metrolist_strings.xml file will be used.

sadorowo:
> > So I decided to create an Updater screen. Let me know where I should place the string key - in strings.xml or metrolist_strings.xml? As I can see keys for names of all settings pages are in strings.xml, but maybe these keys were used in the original app
> 
> This is good, and we will also place enable notifications below it. For any new feature or any addition, the metrolist_strings.xml file will be used.

Okay, thank you. I'll push the recent changes in a while. Should I also create a new preference for notification settings, or you are going to implement this in a separate PR/commit?

mostafaalagamy:
> Okay, thank you. I'll push the recent changes in a while. Should I also create a new preference for notification settings, or you are going to implement this in a separate PR?

If you are going to implement the notifications feature, it's better to include it in this PR. If you leave it for me to implement, I will do it in a separate commit after merging this PR!

sadorowo:
I don't think that I'm experienced enough to implement this, so I'll leave that for you :)

mostafaalagamy:
> I don't think that I'm experienced enough to implement this, so I'll leave that for you :)

Thank you, I will do that!

PR #1664, opened 2025-09-26
refactor: simplify account assignment

FineFindus:
As the values are set to 'Guest'/`null` in either case, we can simplify the assignment to always set the values first and only overwrite them if we have better values. Also removes the now unused method that previously handled the account updates.


mostafaalagamy:
Thank you for reviewing and improving this!

PR #1697, opened 2025-09-28
Fix: Apply audio normalization immediately on track transition

MetalDroid:
This prevents a delay until audio normalization is applied, which caused volume jumps.


mostafaalagamy:
I deleted the branch because it had issues! Will merge it manually.

PR #1707, opened 2025-09-29
New player buttons follow correct md3e button radius and spacing

aayushrautela:
Improve button group layout to match Material Design 3 guidelines

- Reduce spacing between buttons from 12dp to 6dp for tighter grouping
- Adjust corner radius to only reduce inner radius (5dp) while preserving outer radius (50dp) for better visual cohesion
- Applied to both like button group (share/heart) and lyrics mode button group (queue/sleep timer/lyrics/repeat)


mostafaalagamy:
Thank you for your contribution and your time. That's a good catch, I will merge it now!

PR #1719, opened 2025-09-30
feat: Last.fm integration

FullerBread2032:
Added simple scrobbling to Last.fm
This integration send scrobbling information depend on user settings, and also send `Now Playing` message to Last.fm


Zareeff:
> Added simple scrobbling to Last.fm
> This integration send scrobbling information depend on user settings, and also send `Now Playing` message to Last.fm

Works great! Is there a way to make it so when you play music offline (Downloaded) it'll get scrobbled when online just like how metrolist default stat works?
(Sorry for bad english)

FullerBread2032:
Hello!
As for now only listening to song while online, sends scrobble information.
In later updates, I will try to add more services and better integration

PR #1729, opened 2025-10-01
feat(miniPlayer): adjust layout for tablet landscape orientation

aayushrautela:
Adjusts the mini-player layout for tablets (>600dp) in landscape mode, moving it to the bottom-right and reducing its width. This provides a more polished UI on wider screens by preventing the player from awkwardly spanning the full screen.
Old screenshot, I made it a little bigger.
Since we have a separate tablet mode, we can even make it little thicker.
Screenshot_20251001-223731_Metrolist Debug

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/player/MiniPlayer.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/player/MiniPlayer.kt
@@ -250,7 +250,7 @@ private fun NewMiniPlayer(
                 .then(
                     if (isTabletLandscape) {
                         Modifier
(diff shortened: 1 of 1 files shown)


grafspirt:
It will be pain in the ass - to 'catch' the play/pause button somewhere on the right edge of the car screen when driving :( Why don't put the 'button' circle in the center of the screen?

aayushrautela:
> It will be pain in the ass - to 'catch' the play/pause button somewhere on the right edge of the car screen when driving :( Why don't put the 'button' circle in the center of the screen?

It can be made adjustable, but I added it for tablet and right-handed like most other apps.

mostafaalagamy:
Thank you for your contribution and time, I have reviewed it and will merge it now!

PR #1834, opened 2025-10-18
Fix playback loop after call or audio focus changes

MetalDroid:
Adjusted AudioFocus handling to prevent play/pause loops after ending Bluetooth calls by adding reentrancy control and a short delay when regaining focus.


MetalDroid:
Please, don't merge (testing...)

MetalDroid:
Ok, the original bug with Bluetooth devices is fixed (at least in tests with my devices that had the bug).

Anyway, I think that this and other problems would be completely solved by letting the player handle the audio focus:

https://github.com/mostafaalagamy/Metrolist/issues/1833#issuecomment-3419665632

Issue #1835, opened 2025-10-18
More accessible and commonly used actions in "New Design UI"


SageTendo:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

Hi. I’ve been experimenting with some UI ideas for the Player screen, mainly around adding and rearranging controls to make playback options more accessible. 

#### Implementation Thoughts
- Replace or move the existing Share button to make room for Add to Playlist
- Add Shuffle and Repeat icons alongside playback controls
- The sleep timer action moves to the player menu
- The share button can also be found inside the player menu

I'd love to hear your feedback and the community to see if this is something they'd like. And if there are any improvements that could be made :)
If all is good, I can open PRs for the changes for review.

### Why do you want this feature?

Currently, the Player screen doesn’t have include of commonly used actions (shuffle, repeat, or add-to-playlist) directly, which makes it a bit less convenient for quick interaction. I thought these could fit naturally within the new design layout without feeling cluttered.

### Additional information

<img width="200" height="450" alt="Image" src="https://github.com/user-attachments/assets/9e4b6b7a-a23b-4612-bb11-f77ae9d63864" />
<img width="200" height="450" alt="Image" src="https://github.com/user-attachments/assets/62bea8ae-5a04-45e3-8afc-f8d7844c138d" />
<img width="200" height="450" alt="Image" src="https://github.com/user-attachments/assets/2ec08368-888b-463a-8acd-daa1bd747701" />
<img width="200" height="450" alt="Image" src="https://github.com/user-attachments/assets/32b29402-0c28-44ba-bd6e-44a033762794" />

HaleyHalcyon:
Relevant to #1609

grafspirt:
> Currently, the Player screen doesn’t include commonly used actions (shuffle, repeat, or add-to-playlist) directly, which makes it a bit less convenient for quick interaction.

I agree with the 'add-to-playlist' vs 'share' - the 1st one looks as an action you'd need more frequently then sharing. And looks logical next to the 'like' button - you're listening to the track => you like it => you either add it to you favorites, or add it to you playlist, or both. It looks well on the screenshot.

Regarding the two others - I think it was an intention (one of) for the new design - to get rid of such supplementary buttons from the main buttons pannel, to make it a bit more 'lighter'.

[...]

PR #1839, opened 2025-10-19
Feature/nowplaying bottom panel rework

SageTendo:
This PR makes changes to the queue panel on the player screen. The changes are for the new player design only.

#### Changes:
- Replaces the sleep timer button with the share button and places it at the left end of the panel.
- Share button is also added to player menu
- Moves the queue button to the centre of the panel
- Removes the repeat button from the panel (this is added to the player controls in #1838 )
- Moves sleep timer button to the player menu. (This is not changed in the old player design, however the sleep timer can additionally be accessed from the player menu when using the old design as a result of this change)
- **NB:** Moves the sleep timer business logic from the UI and into sleep timer service. This is to allow the sleep timer to be used in the player menu, and the Queue panel for the old UI.

#### Reasons for change:
To better accommodate the design ideas mention in #1835 

Ref: #1835


SageTendo:
An alternative implementation I discussed with another contributor, was to keep the sleep timer at the bottom panel, and then moving the share button to the player menu. Maybe that could be something to look into.

PR #1840, opened 2025-10-19
Feature/nowplaying replace share with playlist

SageTendo:
This PR makes changes to the player screen for the new design

#### Changes:
- Replaces the share button with the "add to playlist" button. The share button would then be moved to the Queue panel in #1839 

#### Reasons for change:
I find having the "add to playlist" button on the player screen is much more convenient as opposed to having to always go to the player menu to access it. It's a frequently used functionality and having one less click would be nice :)

Ref: #1835


grafspirt:
Do you have it here?
Test: You add >1 song to the same playlist,
AR: The system reacts on the 2.. songs (never added yet) as duplicates.

SageTendo:
> Do you have it here? Test: You add >1 song to the same playlist, AR: The system reacts on the 2.. songs (never added yet) as duplicates.

Interesting. I hadn't caught that issue, thanks for that. I couldn't wrap my head around the cause for some time, I think I may have found the problem though. I'll create a fix soon.

SageTendo:
@grafspirt Do you happen to have an idea on why the lambda function is called only if it's null? https://github.com/mostafaalagamy/Metrolist/blob/0a1c4b569abc049c19a34ecec256ed6ff7575dca/app/src/main/kotlin/com/metrolist/music/ui/menu/AddToPlaylistDialog.kt#L107-L109

From my understanding this is fine when the dialog is being called from context menus (since they are recomposed and songIds will always be reinitialised, correct?). But when opening the dialog from the player, the songIds doesn't get updated after the first time it's set.

Shouldn't the songIds be refetched whenever the action to add to the playlist is made? That way it always working with the latest song information.

grafspirt:
Nope.
And this strange remark in the comment:
https://github.com/mostafaalagamy/Metrolist/blob/0a1c4b569abc049c19a34ecec256ed6ff7575dca/app/src/main/kotlin/com/metrolist/music/ui/menu/AddToPlaylistDialog.kt#L67-L69

But all invocations of the 'AddToPlaylistDialog' function are from different menus, so it is supposed to be always with the songIds == null for them.

So in order not to have a possible regression here I suggest to invoke it as (by analogy with some others such dialogs invokation):
`if (showChoosePlaylistDialog) AddToPlaylistDialog(`
`        isVisible = true,`
`        onGetSong = { playlist -> ...`

But I'm not an expert in the Android development so cannot say if it brings any drawback from, e.g., memory usage persepective.

SageTendo:
> So in order not to have a possible regression here I suggest to invoke it as (by analogy with some others such dialogs invokation):

I went with this instead. It seems to be working as intended now, hopefully.

PR #1901, opened 2025-11-04
Fix music service crash (Unknown Error)

adrielGGmotion:
Fix: Defer MusicService property initialization to onCreate

Moves the initialization of context-dependent properties (`audioQuality` and `playerVolume`) from the `MusicService` constructor to the `onCreate()` method.

This prevents a race condition where the `Context` could be `null` during construction, which was causing a `NullPointerException` and crashing the app on startup for some users. The properties are now declared as `lateinit var` and initialized safely within the `onCreate()` lifecycle callback.


adrielGGmotion:
it wasn't supposed to add the other stuff 😂

Issue #1086, opened 2025-07-25
Add a setting to romanize Chinese lyrics


RealNath:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

There are options to romanize Japanese and Korean lyrics. Why not add Chinese too?

### Why do you want this feature?

It could also be useful for people who's learning Chinese

Monoxide8969:
Implemented.

@mostafaalagamy you can close this

PR #1911, opened 2025-11-07
feat: add Chinese lyrics romanization (supports Simplified and Traditional Chinese)

RealNath:
I've added an option to romanize Chinese lyrics. I have no idea why one of the most spoken languages has missing romanization in the first place, even after creating #1086 months ago. I used TinyPinyin for this as it's literally one of the only two Chinese to pinyin Kotlin library I can find. The other one being pinyin4j, but it seems very outdated and has bigger file size compared to TinyPinyin.

There are no tones above the romanized letter (diacritics) for now, because TinyPinyin doesn't support it, and the characters when sung in Chinese songs are different than their original tones when spoken normally anyways.

WhatsApp Image 2025-11-08 at 01 50 08_862cbdd2
WhatsApp Image 2025-11-08 at 01 50 08_a7a7089e


mostafaalagamy:
Thank you for your contribution and your time!

PR #1920, opened 2025-11-11
Add hide video songs option and filter playback

Ughur:
Adds a “Hide video songs” toggle in Content settings so users who only want audio tracks can turn off videos entirely.


app/src/main/res/values/strings.xml
mostafaalagamy, line 1: This should be deleted and added to metrolist_strings

mostafaalagamy:
Thank you for your contribution and your time, I appreciate your effort!

PR #2042, opened 2025-12-20
feat: Add foss/gms build flavors for Google Cast (F-Droid compatible)

### LINKED ISSUE #1115: Cast for Google devices
### Checklist

### Feature description

Integrating an option to cast audio to local supported devices on the network, similar to integrations in Musicolet or the Spotify app. 

By playing music only via bluetooth connection or casting the screen from the Google home app, a user is unable to play music in the background while listening to music from other sources, such as playing music for children while listening to an audiobook through bluetooth. 

### Additional information

_No response_

liraz747:
Closes #1115

Follows up on the original Cast PR that was merged - this adds build flavor support so Cast works on F-Droid too

### What this does
Separates Google Cast into a `gms` build flavor so the default build (`foss`) does not include any Google Play Services dependencies. This addresses the F-Droid compatibility concern raised by @FineFindus.

### How it works
- `foss` flavor (default): No Cast functionality, no GMS libs - this is what F-Droid will build
- `gms` flavor: Full Cast support like before

The `foss` flavor has stub implementations so the code compiles without needing the Cast libs. The Cast toggle in Settings only shows up in `gms` builds.

### Changes
Build config:
- Added `variant` flavor dimension alongside existing `abi` dimension
- Cast dependencies now use `gmsImplementation` instead of `implementation`
- `foss` is set as default so existing CI commands work unchanged

New source sets:
- `app/src/foss/` - stub files (empty composables, no-op handlers)
- `app/src/gms/` - moved the real Cast implementation here

Modified:
- PlayerSettings checks `BuildConfig.CAST_AVAILABLE` before showing Cast toggle
- MusicService works with both stubs and real implementation

### Testing
Both `assembleUniversalFossDebug` and `assembleUniversalGmsDebug` build and run correctly. FOSS build has no Cast UI, GMS build works like before.


mostafaalagamy:
Very good, thank you for your interest in this as well. I will merge it for testing!

FineFindus:
Thanks, this is much appreciated.

PR #2061, opened 2025-12-24
FEAT: implement shuffle playlist first feature in MusicService + fix retry on source error

FrancescoGrazioso:
- Added a new preference key for shuffling original playlist items before similar content.
- Updated MusicService to handle the new shuffle logic based on user preference.
- Enhanced PlayerSettings UI to include a toggle for the shuffle playlist first option.
- Added corresponding string resources for the new feature in both English and Italian.

This feature improves user control over playback order during shuffling.

This feature also aim to fix some inconsistencies in the source error fix, right now it correctly retry every 3 seconds but if the link is expired it will be blocked in an endless loop.


app/src/main/res/values/strings.xml
mostafaalagamy: This also needs to be reverted because it translates through Weblate!

mostafaalagamy:
Thank you for your contribution, I will merge it now!

Issue #1772, opened 2025-10-06
Lyrics Romanizes already Latin characters


MegaMato17:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

So, some some songs when you look at the lyrics, they're Romanized, but the original songs is still English

### Expected behavior

Lyrics shouldn't need Romanized, 

### Actual behavior

it just has the same text underneath it

### Screenshots/Screen recordings

![Image](https://github.com/user-attachments/assets/0252dcd8-ee5e-48b5-8d01-c0a90aba5c3a)

### Logs

N/A

### Metrolist version

12.7.0

### Android version

Android 16

itsmeteddyhere:
I have noticed this as well, it seems to be the Cyrillic romanization causing it, I don't know what exactly though. I'll look into it, in the meantime you can turn off the "romanize current track" toggle for the tracks it happens on.

FullerBread2032:
This is due to the fact that the lyrics most likely contain the Cyrillic letter `e` instead of the usual
Which has `U+0435` code instead of `U+0065`.
This is mostly issue with lyrics provider, but maybe romanization check should work better and catch this

itsmeteddyhere:
Yeah, I'll take a look and try to fix it.

Dr-Brixx:
Are any of you still experiencing that issue?

MegaMato17:
> Are any of you still experiencing that issue?

Depending on the song, yeah

Dr-Brixx:
That's bad. Can you give me a song example and the lyrics provider used?

Dr-Brixx:
And does turning on detect language line by line help?

nyxiereal:
Closing as a duplicate of #3638. Latin text being unnecessarily romanized is covered by the canonical romanization issue.

Please continue discussion and add any current reproduction details on #3638.

PR #2176, opened 2026-01-01
Already-latin lyrics containing cyrillic e don't get romanized anymore

itsmeteddyhere:
I have added a check to see if the only character is the cyrillic e, incase it is it just doesn't romanize it. This serves as a fix for #1772


mostafaalagamy:
Welcome back!

PR #2193, opened 2026-01-02
feat: Add Full Parametric EQ with Custom DSP

ndellagrotte:
This PR replaces Metrolist's existing equalizer with a custom DSP implementation supporting full parametric EQ functionality, ported from my own local music player [NanoSonic](https://github.com/ndellagrotte/NanoSonic).

Metrolist's current equalizer implementation uses android.media.audiofx.Equalizer, which is limited to a fixed 5-band graphic EQ. My proposed replacement implementation would enable users to import and apply device-specific correction profiles from [AutoEQ](https://autoeq.app/).

### Key Changes

- Replaced android.media.audiofx.Equalizer with custom DSP implementation
- Supports arbitrary number of EQ bands (no fixed limit)
- Full parametric control: per-band frequency, gain, and Q (bandwidth)
- Added support for filter types: peaking, low-shelf, high-shelf, lowpass, highpass
- Redesigned EQ Screen using Material 3 components
- Added logic for importing of custom EQ profiles via system file picker
- EQ state applied and restored within MusicService lifecycle

### Testing

- Imported and applied 10+ band AutoEQ profiles for various headphones
- Tested profile switching during active playback
- Confirmed EQ profile persists across playback service restarts

### Notes for Reviewers

- The legacy equalizer was not removed from the codebase, however, users can't access it
- New dependency for icons which is definitely not necessary (I got lazy)
- Performance impact with a 10-band PEQ is very low. I've yet to limit-test this
- There is no GUI for modifying or creating your own EQ profile. However, this shouldn't be terribly difficult to implement.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/App.kt
+++ b/app/src/main/kotlin/com/metrolist/music/App.kt
@@ -17,6 +17,7 @@ import coil3.PlatformContext
 import coil3.SingletonImageLoader
 import coil3.disk.DiskCache
 import coil3.disk.directory
+import coil3.memory.MemoryCache
 import coil3.request.CachePolicy
 import coil3.request.allowHardware
 import coil3.request.crossfade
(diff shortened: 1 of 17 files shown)


Zareeff:
This only works on Android 13 or above, right?

GoldenWarriorM:
Works on android 8.1 x86_64 emulator
image

mostafaalagamy:
This is truly impressive, thank you for your time and contribution, and I look forward to working with you again! I will merge it now!

mostafaalagamy:
Could you write some instructions about this, or how to use it? It's possible to add it to the README file!

ndellagrotte:
Sure. I'll PR something, but in the future, there should really be some kind of function that fetches the AutoEQ profiles automatically. However, the process of obtaining one yourself is relatively simple. You'd just have to use the AutoEQ webapp for now.

PR #2221, opened 2026-01-04
fix: playerBottomSheet slow/laggy animations

ndellagrotte:
- This PR fixes a performance regression involving playerBottomSheet's expansion/contraction animation
- I fixed this by turning the equalizer screen into a dialog, thereby bypassing the need for the stupid navbar logic that caused this problem in the first place
- Equalizer screen looks and works better as a dialog
Screenshot_20260103_233613


ndellagrotte:
erm

PR #2227, opened 2026-01-04
fix/refactor/UI: Fix playerBottomSheet's buggy animation, related UI changes and refactoring

ndellagrotte:
### Fixes
- Removed faulty navbar logic responsible for a performance regression involving playerBottomSheet's expansion/contraction animation
- Removed `collapseSoft()` call from equalizer menu item's `onClick` handler, which was previously causing the player to minimize after the equalizer menu had been accessed 

### UI
- Converted `EqScreenContent()` into a dialog
- EQ Profile import success message snackbar removed, and its message is instead silently logged with timber
- The corresponding error message is also logged with timber and notifies the user without using a snackbar
- Adjusted and simplified display text in `EqScreenContent()`

### Refactoring
- Refactored `MainActivity` by extracting the entire UI tree into a dedicated `@Composable` function named `MetrolistApp`

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/equalizer/EqScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/equalizer/EqScreen.kt
@@ -1,5 +1,6 @@
 package com.metrolist.music.ui.screens.equalizer
 
+import android.annotation.SuppressLint
 import android.media.session.PlaybackState
 import android.provider.OpenableColumns
 import androidx.activity.compose.rememberLauncherForActivityResult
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/com/metrolist/music/ui/screens/equalizer/EqScreen.kt
mostafaalagamy, line 1: The text should be converted to a string, like this: "text = "Equalizer","

mostafaalagamy:
Okay, I think it's ready now. I'll merge it to test it in everyday use!

mostafaalagamy:
I cleaned the code!

PR #2248, opened 2026-01-06
Order playlists by name when adding song

jimmy-sh:
fix mostafaalagamy/Metrolist#1407
orders playlists alphabetically in add to playlist dialog


mostafaalagamy:
Thank you for your time and your initial contribution. I hope we can always work together!

Issue #1407, opened 2025-08-24
Sorting options for "Add to playlist" pop-up window


Username1-a:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

When adding a song to a playlist, after clicking "Add to playlist" button, pop-up window appears listing existing playlists

Playlists in this pop-up are sorted only from most recently created at the top of the list to the oldest playlist at the bottom

It would be great if there was several sorting options in this pop-up, primarily just "Alphabetical" (from A to Z, ascending) (other options are also appreciated)

### Why do you want this feature?

It would greatly speed up the process of adding a song to a playlist since if there's a lot of playlists in the library it can be very cumbersome to find a particular playlist in that pop-up window since they're all usually created at different times so there's mishmash of completely different playlist names in the list — making sorting here alphabetical by default definitely gonna streamline it
 
Sorting in this window gets overlooked in all InnerTune forks unfortunately, so could be a nice addition to Metrolist

jimmy-sh:
Hello! I have made a fix for this so the playlists appear ordered by name
How could i push them to this repo?
@mostafaalagamy , could you please help me on getting the rights?

<img width="599" height="126" alt="Image" src="https://github.com/user-attachments/assets/54e08ee2-ccba-4132-a926-e7e549ce5f86" />

mostafaalagamy:
> Hello! I have made a fix for this so the playlists appear ordered by name
> How could i push them to this repo?
> @mostafaalagamy , could you please help me on getting the rights?
> 
> <img width="599" height="126" alt="Image" src="https://github.com/user-attachments/assets/54e08ee2-ccba-4132-a926-e7e549ce5f86" />

Push it in your github fork, then perform a comparison and open the PR!

jimmy-sh:
> Push it in your github fork, then perform a comparison and open the PR!

Did so, thanks. First time of such a contribution 😄

jimmy-sh:
Hello!
I've just PR'd a commit that adds sorting

PR #2257, opened 2026-01-07
Add playlist sorting when adding song to playlist

jimmy-sh:
close mostafaalagamy/Metrolist#1407


jimmy-sh:
This is the commit that covers the sorting part of issue #1407 
It uses the library's playlist sort

PR #2258, opened 2026-01-07
style: remove NewActionButton elevation?

GoldenWarriorM:
I think current NewActionButton design with shadow looks out of place since other buttons don't have any shadow. Maybe some outline should be added for buttons to pop out a bit more (like with mini player)? To be honest on phone it looks a lot worse then in screenshots from vm (probably because no antialiasing in vm).
 
 WIth elevation (current):
 
image

 No elevation:
image


GoldenWarriorM:
A real device with a light theme shadow looks like this
image

PR #2303, opened 2026-01-11
Play after first play command on widget

jimmy-sh:
After app stopped, play song after first play command on widget, if a song is already in queue, even if "No song playing" is displayed. Before this change, it loaded the song but an additional tap on play would be necessary

fix mostafaalagamy/Metrolist#2300


mostafaalagamy:
https://github.com/mostafaalagamy/Metrolist/actions/runs/20899470944

Try this build, the widget design logic has been completely changed!

Contact me on Telegram so I tell you some things if you can!

PR #2333, opened 2026-01-12
Feat: Add faster silence skipping

### LINKED ISSUE #2322: Silence insta-skip
### Checklist

### Feature description

I don't know if it's a good idea or if it's possible to make, but I had this in mind since I was listening to MX by Deftones, which track in all versions of the album Around The Fur has a +30 minutes of silence with a secret song in it. I tried putting the Skip Silence function on and I realized it wasn't as I expected to be, because it doesn't SKIP itself, it just turns on a higher speed until the song has a sound again (which is really slow in this case), so I originally thought it was about to jump suddenly to the exact minute the song starts the second part. I really think changing the function of that feature to a quick jump (or having it as optional) would be really useful.

I hope this suggestion is taken into account. Cheers!

Image 

Comfort

### Additional information

_No response_

nyxiereal:
The current "Skip silence" option just speeds up the currently playing music until the silence ends. I added an "Instantly skip silence" option that hops until it finds audio. I find it way faster than just speeding up the song from my testing.

I also clarified both mentioned settings in player and audio options, so its more clear what they do.

Nevermind, just use the build made by the gh workflow.

Closes #2322

### DIFF
--- a/app/src/main/res/values/metrolist_strings.xml
+++ b/app/src/main/res/values/metrolist_strings.xml
@@ -133,6 +133,9 @@
     <string name="shuffle_playlist_first">Shuffle playlist/album first</string>
     <string name="shuffle_playlist_first_desc">When shuffling, play all songs from the original playlist/album first, then similar content</string>
     <string name="show_wrapped_card">Show Wrapped card</string>
+    <string name="skip_silence_desc">Fast forward through silent parts of songs</string>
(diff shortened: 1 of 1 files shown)


mostafaalagamy:
I have made some changes to the strings, thank you for your contribution and your time, and I look forward to more from you!

nyxiereal:
Ah, well it's my first time working on a project with translations. I'm definitely gonna contribute more!

Issue #2211, opened 2026-01-03
Add artists' info / bio to their page


Monoxide8969:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

Have a small section dedicated to artists' bio pulled from wikipedia on the bottom of their page.

The section should be expandable on tap to avoid unrequested walls of text

### Why do you want this feature?

Nice insight on the artists, gives more context about them

### Additional information

Other apps such as kreate or vitune already do this

![Image](https://github.com/user-attachments/assets/4ae04fa0-8cbe-4d4f-b031-239e001e9e38)

nyxiereal:
<img width="1440" height="1485" alt="Image" src="https://github.com/user-attachments/assets/ebb8c8ba-d276-446d-bada-3ed46a51151f" />

Is this what you'd like?

Monoxide8969:
It's exactly it

nyxiereal:
I'll polish it up a bit and fix things up tomorrow. I'll make a pull request after that.

nyxiereal:
<img width="1440" height="2497" alt="Image" src="https://github.com/user-attachments/assets/16759c3c-f28a-496c-ad20-89485b0e0fb8" />

Update

Monoxide8969:
Amazing, the subs and the audience is another nice detail to add.

Thanks a lot, can't wait for it to be merged

PR #2348, opened 2026-01-13
Feat: Add more artist info

nyxiereal:
this PR adds the ability to view artist's description, subscriber count, monthly listeners. these can be toggled from the content settings under "Artist page". the description text is parsed in the same way as ytm's desktop app, so it'll work just fine. subscribers aren't visible in the desktop app, monthly listeners sometimes aren't. but that's just a quirk of ytm i guess.

image
image


mostafaalagamy:
Good work!

nyxiereal:
Oh, and this fixes #2211

PR #2438, opened 2026-01-21
fix(VolumeSlider): hide circle if the volume is more than 90%

omardotdev:
this hides the circle (at the end of the slider) when the volume is more than 90% so it would look something like this
IMG_20260121_201538
instead of this
IMG_20260121_144721


nyxiereal:
LGTM

PR #2445, opened 2026-01-22
Update strings.xml

dpentx:
Hello!
I tried to fix a forgotten string line in the media player UI.
It looks like Weblate did not include some strings properly.
Because of that, the Pause / Play button text is incorrect:
when the music is paused, it shows "Çal" instead of "Pause", and vice versa.
This PR fixes the missing / wrong string mapping so the correct text is shown depending on the playback state.
Thanks 🙌
Screenshot_20260122_221549_Metrolist
Screenshot_20260122_221545_Metrolist

### DIFF
--- a/app/src/main/res/values-tr/strings.xml
+++ b/app/src/main/res/values-tr/strings.xml
@@ -50,7 +50,9 @@
     <string name="details">Detaylar</string>
     <string name="edit">Düzenle</string>
     <string name="start_radio">Radyo başlat</string>
-    <string name="play">Çal</string>
+    <string name="play">Oynat</string>
+    <string name="pause">Duraklat</string>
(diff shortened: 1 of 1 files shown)


nyxiereal:
> \> Task :app:mergeUniversalFossDebugResources FAILED
ERROR: [string-tr/play_pause] /home/runner/work/Metrolist/Metrolist/app/src/main/res/values-tr/strings.xml [string-tr/play_pause] /home/runner/work/Metrolist/Metrolist/app/src/main/res/values-tr/metrolist_strings.xml: Resource and asset merger: Duplicate resources

nyxiereal:
please use weblate for submitting translations. also, only the `play` one is in the `strings.xml` file, the rest is in `metrolist_strings.xml`. please edit that file or do your edits on weblate

PR #2452, opened 2026-01-23
feat(settings): Keep screen on when player is expanded

tashilapathum:
I liked this behaviour from some other music players. But since I'm not sure if everyone would like it, I added it as a setting under **Player and audio -> Misc**.

screenshot

I couldn't think of an icon to use so I reused an already avaiable one. Hope it looks alright for the setting.


app/src/main/res/values/strings.xml
mostafaalagamy, line 1: Revert this and apply it within metrolist_string

tashilapathum: Done

PR #2455, opened 2026-01-23
feat(settings): Crop music video thumbnails

tashilapathum:
This PR adds a new setting under **Appearance -> Player** to enable cropping music video thumbnails, so their size looks consitent with regular album arts. 

It might just be my personal preference but I think this looks neat because:
- This makes use of the empty vertical space of expanded player when playing music videos
- Looks consistent with the mini player album art thumbnail
- Focuses the subjects of most of the thumbnails
- The user can still see the full thumbnail on home screen to identify music videos at a glance
- Some thumbnails look kinda artisitic zoomed in :D

https://github.com/user-attachments/assets/4194c831-d5b3-4199-918d-00d176bb5e28

Some thumbnails do look weird when cropped tho.


mostafaalagamy:
Thank you for your contribution and your time, I will merge it now!

PR #2521, opened 2026-01-29
Move hardcoded buttons text to strings

123jjck:


app/src/main/res/values/strings.xml
mostafaalagamy: It must be inside the metrolist_strings, Then use the translation from Weblate!

app/src/main/res/values-ru/strings.xml
mostafaalagamy: Revert

123jjck:
I just noticed that there are also `Mute` strings hardcoded here

mostafaalagamy:
Thank you for your contribution and your time, I will merge it now!

PR #2565, opened 2026-01-31
fix: handle 403 errors when getting player metadata

suvojeet-sengupta:
Had to patch YTPlayerUtils cause the main client (Android VR I think?) was throwing 403s for some people.
Added a fallback loop so if the first request fails, it just tries the other clients (like web remix or ios) instead of crashing immediately.
Also made sure we don't fetch the tracking info twice if we already fell back to the web client.

Should fix that IO_BAD_HTTP_STATUS error report.


suvojeet-sengupta:
Please build it fast bro

suvojeet-sengupta:
Need to test

suvojeet-sengupta:
@mostafaalagamy

PR #2686, opened 2026-02-08
Fixes playback issue.

LuckySilver0021:
Resolves for #2556 .

Tries to resolve the play back issue.

The issue was because of :
race condition between Android service binding  & player initialization,

and PlayerConnection created before MusicService.player was ready.

Now :
added readiness flag,
PlayerConnection uses safe accessors with exception handling.
and MainActivity retries binding if creation fails.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/MainActivity.kt
+++ b/app/src/main/kotlin/com/metrolist/music/MainActivity.kt
@@ -15,6 +15,7 @@ import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
 import android.os.IBinder
+import android.util.Log
 import android.view.View
 import android.view.WindowManager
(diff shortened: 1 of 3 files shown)


nyxiereal:
pls use Timber.tag instead of Log.

mostafaalagamy:
Review the fix in the other PR and apply it here if it works for you!

mostafaalagamy:
LaunchedEffect(mediaMetadata?.id) {
        showInlineLyrics = false
}

This needs to be removed, and the conflict resolved!

LuckySilver0021:
> LaunchedEffect(mediaMetadata?.id) { showInlineLyrics = false }
> 
> This needs to be removed, and the conflict resolved!
Yea I got that,
And please dont mind me Im fixing the linting isssues.
Just a moment please

PR #2818, opened 2026-02-14
fixes random ForegroundService crash

AbhinandAK350:
Removes the conditional `startForegroundService` calls in widget receivers, simplifying them to just `startService`. This is safe because the service already calls `startForeground` itself when appropriate.


adrielGGmotion:
@AbhinandAK350 Build faied because of unresolved references, make sure the app builds next time before you create an PR and that your fix indeed works

AbhinandAK350:
Sure

Issue #2990, opened 2026-02-22
Persist default sleep timer duration


johannesbrauer:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

Implement an option to save a user-defined default duration to improve user experience:
* Add a **"Set as Default"** button to the sleep timer dialog.
* Persist the selected value using `DataStore`.
* Automatically initialize the timer with the stored value upon opening the dialog.

### Why do you want this feature?

Currently, the sleep timer resets to its hardcoded initial value every time the dialog is opened. This requires users to manually adjust the duration each time if they prefer a different default value.

### Current Status
I have already implemented this feature locally and verified it within the app. The implementation:
- Uses `DataStore` for persistent storage.
- Follows the existing Material 3 design of the app.
- Includes a Toast confirmation when a new default is saved.

**Note:** Please let me know if you are interested in a Pull Request for this enhancement.

PR #2991, opened 2026-02-22
feat: Add persistent default value for sleep timer

johannesbrauer:
Closes: #2990 

## What
Adds the ability to save a user's preferred sleep timer duration as a persistent default value. When the sleep timer dialog is opened, it automatically loads the previously saved duration instead of resetting to 30 minutes.

## How
- **PreferenceKeys.kt**: Added `SleepTimerDefaultKey` (`floatPreferencesKey`) for DataStore persistence
- **Player.kt & Queue.kt**: 
  - Load initial value from `SleepTimerDefaultKey` via `rememberPreference()` (fallback: 30f)
  - Added "Set as Default" `TextButton` in sleep timer dialog
  - Button saves current slider value to DataStore with Toast confirmation
- **metrolist_strings.xml**: Added `set_as_default` and `sleep_timer_default_set` string resources

## Screenshots
| Sleep Timer Dialog (Before) | Sleep Timer Dialog (After) |
|-----------------------------|----------------------------|
| Screenshot_2026-02-22-22-08-10-154_com metrolist music | Screenshot_2026-02-22-22-08-16-077_com metrolist music debug |

| Before | After |
|--------|-------|
| Dialog always resets to 30 min | Dialog loads saved default value |
| No "Set as Default" option | New button allows saving preference |

## Checklist
- [x] Follows existing code style
- [x] Both Player and Queue dialogs updated for consistency
- [x] DataStore integration matches existing preference patterns
- [x] String resources added to metrolist_strings.xml
- [x] No breaking changes

### DIFF
--- a/app/src/main/res/values/metrolist_strings.xml
+++ b/app/src/main/res/values/metrolist_strings.xml
@@ -255,6 +255,10 @@
         <item quantity="other">%d seconds</item>
     </plurals>
 
+    <!-- Sleep timer -->
+    <string name="set_as_default">Set as default</string>
+    <string name="sleep_timer_default_set">Sleep timer default set to %d min</string>
+
     <!-- Player Settings -->
     <string name="disable_load_more_when_repeat_all">Disable load more when repeat all</string>
     <string name="disable_load_more_when_repeat_all_desc">Don\'t auto load more songs and similar content when repeat all mode is enabled</string>


nyxiereal:
Use `metrolist_strings.xml`, not `strings.xml` for your strings. Other than that LGTM

johannesbrauer:
Got it, I'll update the PR as soon as possible.

nyxiereal:
remove the strings you added to `strings.xml` and move them to `metrolist_strings.xml`

johannesbrauer:
Ok but I did already, it was only two strings, I removed them from strings.xml and pastet them to metrolist_strings.xml. Or did you wanted that as a separate commit?

johannesbrauer:
On GitHub desktop you can click on the forcepushed and there you can see that it's moved.

nyxiereal:
Oh I just didn't see that in the web interface

PR #3012, opened 2026-02-24
Fix: Restore v13.0 stable playback logic 

SykikXO:
Fix: Restore v13.0 stable playback logic and version bump v13.1.2

## Description
This PR addresses critical playback issues encountered in v13.1.x and restores the stable playback logic from the v13.0 baseline.
Got back the NewPipe logic from v13.0 and it fixed the issue, i think the streamlining logic should be delayed as making that main currently bricks cache, i have over 700 songs (backup) and would hate to lose them (still lost about 200). Also had to tweak the build memory to get it to compile on my end. 

Also, bumped the version to 13.1.2 for my own fork testing,please do as you wish with that (im sorry).  

Verified the fix by building a full self-signed Release (FOSS variant). Playback is stable on multiple tracks that were previously failing with 'Video unavailable' or IO errors. Tested on [OnePlus CPH2719 / Android 16].

### Key Fixes:
1. **Restored Stable Playback Logic**: Re-enabled `YouTube.newPipePlayer` robust stream extraction fallback, which was the baseline for stable playback in v13.0.
2. **Refined Deobfuscation Flow**: Refined the application order of n-transforms and PoTokens to ensure consistent stream hydration across various YouTube clients.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/cipher/CipherWebView.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/cipher/CipherWebView.kt
@@ -16,11 +16,11 @@ import timber.log.Timber
 
 class CipherWebView
 private constructor(
-        context: Context,
-        private val playerJs: String,
-        private val sigInfo: FunctionNameExtractor.SigFunctionInfo?,
-        private val nFuncInfo: FunctionNameExtractor.NFunctionInfo?,
-        private val initContinuation: Continuation<CipherWebView>,
+    context: Context,
+    private val playerJs: String,
+    private val sigInfo: FunctionNameExtractor.SigFunctionInfo?,
+    private val nFuncInfo: FunctionNameExtractor.NFunctionInfo?,
+    private val initContinuation: Continuation<CipherWebView>,
 ) {
     private val webView = WebView(context)
     private var sigContinuation: Continuation<String>? = null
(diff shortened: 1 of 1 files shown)


SykikXO:
i have a self signed foss release .apk at  https://litter.catbox.moe/eew74aj8zmitrkxh.apk

nyxiereal:
You are absolutely not permitted to bump the version yourself, who the fuck do you think you are?

alltechdev:
Hey, appreciate the effort on this PR!

A few things I noticed though:

The newPipePlayer method was never actually removed - it existed in the codebase but wasn't being called (there's a comment that says "Skip NewPipe"). So this is more about activating it rather than restoring it.

The main issue is that NewPipe isn't actually doing anything. I ran logcat while playing tracks and here's what shows up:

```
YTPlayerUtils: Finding stream URL for format: audio/webm; codecs="opus", videoId: oZTtJIQdyeM, skipNewPipe: false
Metrolist_CipherWebView: Signature deobfuscated: AJEij0EwRQIhAP_a-xXHjpVZoUjkKa...
Metrolist_CipherDeobfusc: Creating CipherWebView with sig=Kb, constantArg=21, nFunc=null[null]
Metrolist_CipherWebView: Signature deobfuscated: gpSvxCy1R43QuPbMQ_DRXTRkSLhEjg...
```

NewPipe gets called but there's no success log from it - CipherWebView handles all the actual deobfuscation. So the NewPipe call is just adding latency before falling back to what's actually working.

Also you've got NewPipe set as primary with the custom cipher as fallback - that's backwards since NewPipe isn't functional right now.

The CipherWebView changes also have some issues:
- `window.onerror = function() { return true; }` suppresses ALL JS errors which could hide real problems
- `if (key.length > 3) continue` is too aggressive. Look at the logs: `N-function NOT found (brute_force:tested=1/1119)` - only 1 function tested out of 1119.
- The try/catch around fn(testInput) is fine though

Other stuff:
- The build config changes (LTO off, 8GB memory, config cache disabled) look like local tweaks - probably better to keep those out of upstream
- Same with the version bump if it was just for fork testing

---

**Update:** I did some more testing and the `key.length > 3` filter is actually breaking WEB_REMIX entirely.

Here's main branch playing a track:
```
Metrolist_CipherWebView: N-function NOT found (brute_force:tested=184/1119)
Metrolist_EjsNSolver: N-transform result: VKro05EImdeU8Q
YTPlayerUtils: N-transform applied successfully
YTPlayerUtils: Stream URL validation result: Success (200)
YTPlayerUtils: Playback: client=WEB_REMIX
```

And here's your PR branch:
```
Metrolist_CipherWebView: N-function NOT found (brute_force:tested=1/1119)
YTPlayerUtils: Stream URL validation result: Failed (403)
Metrolist_CipherDeobfusc: N-transform function was not discovered at init time
YTPlayerUtils: Trying fallback client 1/11: TVHTML5_SIMPLY_EMBEDDED_PLAYER
YTPlayerUtils: Playback: client=TVHTML5
```

See the difference? Main tests 184 functions, your PR tests only 1 because of that filter. The n-transform fails, WEB_REMIX gets a 403, and it falls back to TVHTML5.

So what's actually happening is your CipherWebView change broke WEB_REMIX, which forces the fallback to TVHTML5. TVHTML5 gives direct URLs so it doesn't need the cipher. That's why playback "works" on your build - not because of NewPipe, but because of the fallback to a client that doesn't need the cipher at all.

This would remove access to WEB_REMIX, our primary highest quality client.

adrielGGmotion:
> You are absolutely not permitted to bump the version yourself, who the fuck do you think you are?

Holy shit they're mad

SykikXO:
@nyxiereal Im really sorry for the version bump thing, was testing with that and i thought for successful installation as an update to carry over cache it was required, i didnt mean to be rude. 

@alltechdev i understand, i thought that playback threw error whenever i tried playing a liked song that wasnt cached but imported from a previous backup, so i figured i would jump to how it used to work in v13. i need to spend more time with logs i guess before creating a pr. thankyou for review, sorry i wasted your time. 

sorry @nyxiereal , im new to this. i made a mess and im ashamed of it.

SykikXO:
> > You are absolutely not permitted to bump the version yourself, who the fuck do you think you are?
> 
> Holy shit they're mad

They have every right to be, im dumb.

alltechdev:
> @nyxiereal Im really sorry for the version bump thing, was testing with that and i thought for successful installation as an update to carry over cache it was required, i didnt mean to be rude.
> 
> @alltechdev i understand, i thought that playback threw error whenever i tried playing a liked song that wasnt cached but imported from a previous backup, so i figured i would jump to how it used to work in v13. i need to spend more time with logs i guess before creating a pr. thankyou for review, sorry i wasted your time.
> 
> sorry @nyxiereal , im new to this. i made a mess and im ashamed of it.

Yeah... Gotta figure out the backup thing regardless. 

Thanks for trying :)

PR #3049, opened 2026-02-26
fix: n-transform extraction for uploaded/restricted songs

alltechdev:
## Problem
Uploaded songs (MUSIC_VIDEO_TYPE_PRIVATELY_OWNED_TRACK), age-restricted songs, and some restricted content return HTTP 403 when streaming. The n-parameter in stream URLs is not being transformed correctly.

## Cause
YouTube changed their player.js structure in Feb 2026. The n-transform function is now stored in an array pattern:
```javascript
var v4c=[yzP];  // Function stored in array
e=v4c[0](e),f[M[42]](M[26],e)  // Called as array[0](param)
```

The `EjsNTransformSolver` uses bundled JS files (`yt.solver.core.js`, `meriyah.js`, `astring.js`) to preprocess player.js and extract the n-function. This preprocessor cannot handle the new obfuscation pattern and fails silently. Fixing it would require updating the bundled `yt.solver.core.js` script.

## Solution
- Switch from `EjsNTransformSolver` to `CipherDeobfuscator` for n-param transformation (uses WebView with function export/brute-force discovery)
- Add exact context pattern to `FunctionNameExtractor` that matches the specific n-transform call site: `e=ARRAY[0](e),f[M[42]]`
- This pattern avoids false positives by requiring the same variable on both sides and the M[42] (set) operation

## Testing
- Built and installed universalFossDebug APK
- Played uploaded songs that were previously returning 403
- Tested age-restricted content
- Verified n-transform logs show correct array discovery (v4c[0])
- Streams return 200 instead of 403

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
@@ -252,7 +252,7 @@ object YTPlayerUtils {
                 if (needsNTransform) {
                     try {
                         Timber.tag(logTag).d("Applying n-transform to stream URL for ${currentClient.clientName}")
-                        streamUrl = EjsNTransformSolver.transformNParamInUrl(streamUrl!!)
+                        streamUrl = CipherDeobfuscator.transformNParamInUrl(streamUrl!!)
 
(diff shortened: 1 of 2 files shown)


mostafaalagamy:
This's closed because it was merged into the last commit

Issue #3057, opened 2026-02-27
Make "Stable Volume" (Audio Normalization) feature turned off by default


IvoryIlya:
### Checklist

- [x] I've checked that there is no other issue about this feature request.
- [x] This issue contains only one feature request.
- [x] The title of this issue accurately describes the feature request.

### Feature description

Please make "Stable Volume" (Audio Normalization) option get turned off by default

### Why do you want this feature?

It harms the the spatiality of music (https://youtu.be/QApKypVHepE)

### Additional information

~I can also make a PR (probably) if you do want to implement this~
Update: made a PR, passed the checks

nyxiereal:
not intended, any user can just disable it in settings

PR #3058, opened 2026-02-27
Switched Audio Normalization (Stable Volume) off by default

IvoryIlya:
## Problem
Audio Normalization (Stable Volume) is on by default

## Solution
- I have changed the defaultValue of audioNormalization to false.

References this issue created by me. #3057


nyxiereal:
not intended, any user can just disable it in settings

PR #3106, opened 2026-03-02
feat: save and restore podcast episode playback position

alltechdev:
## Summary
- Save podcast episode playback progress locally
- Restore position when returning to an episode
- Position saved on pause, track transition, every 15 seconds, and on service destroy
- Only saves after 3 seconds of playback to avoid trivial positions
- Database migration 34 → 35 adds `playbackPosition` column to song table


alltechdev:
To be clear, this is after switching to another episode, and THEN returning to previous.

Old behavior is to start from the beginning.

PR #3120, opened 2026-03-02
fix: uploaded and restricted track playback with WebView-based PoToken

alltechdev:
## Summary
- Restore WebView-based PoToken and EJS cipher implementation for reliable playback
- Revert n-transform optimization that broke WEB_REMIX playback
- Restore YouTubeClient version updates and TVHTML5 potoken flag

## Changes
- Revert to WebView-based cipher deobfuscation which is more reliable
- Regular tracks use WEB_REMIX with proper PoToken authentication
- Uploaded/private tracks use TVHTML5 client with validation skipped
- Explicit/restricted tracks fall back to TVHTML5 when WEB_REMIX returns 403

## Test plan
- [x] Regular tracks play via WEB_REMIX client
- [x] Uploaded/private tracks play via TVHTML5 client (validation skipped)
- [x] Explicit/restricted tracks fall back to TVHTML5 on 403

### DIFF
--- a/innertube/src/main/kotlin/com/metrolist/innertube/models/YouTubeClient.kt
+++ b/innertube/src/main/kotlin/com/metrolist/innertube/models/YouTubeClient.kt
@@ -50,14 +50,14 @@ data class YouTubeClient(
 
         val WEB = YouTubeClient(
             clientName = "WEB",
-            clientVersion = "2.20260124.00.00",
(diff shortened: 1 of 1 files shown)


alltechdev:
For anyone wondering, it's because YouTube tightened their restrictions on uploaded and restricted playback sometime in the last 24hrs.

Without botguard no longer working.

PR #3141, opened 2026-03-03
feat: make cached songs globally available

kairosci:
## Problem
Cached songs are currently only available within the cache folder and are not treated as offline songs throughout the app.

## Cause
The existing implementation strictly checks for the `isDownloaded` status in the database to identify offline songs, ignoring songs that are present in the player cache but were not explicitly downloaded.

## Solution
- Added `isCached` column to `SongEntity` in the Room database (incremented DB version and added migration).
- Created `MetrolistCacheEvictor.kt` to hook into ExoPlayer cache events and synchronize the `isCached` status to the database.
- Added initialization block in `MusicService` to perform initial cache sync.
- Updated database DAO queries to include songs where `isCached = true` alongside `isDownloaded = true` for offline playlists.
- Removed legacy dedicated "Cached Playlist" UI separation to unify the offline experience.

## Testing
- Successfully compiled the project.
- Verified ExoPlayer API correctly tracks spans and cache status and synchronizes with the database.
- Tested the app with previously saved songs.

## Related Issues
- None

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryMixScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryMixScreen.kt
@@ -59,6 +59,7 @@ import com.metrolist.music.constants.MixSortDescendingKey
 import com.metrolist.music.constants.MixSortType
 import com.metrolist.music.constants.MixSortTypeKey
 import com.metrolist.music.constants.MixSortTypeKey
+import com.metrolist.music.constants.ShowCachedPlaylistKey
 import com.metrolist.music.constants.ShowDownloadedPlaylistKey
 import com.metrolist.music.constants.ShowLikedPlaylistKey
 import com.metrolist.music.constants.ShowTopPlaylistKey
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryMixScreen.kt
```diff
@@ -144,16 +144,6 @@ fun LibraryMixScreen(
...
-    val cachePlaylist =
-        Playlist(
-            playlist = PlaylistEntity(
-                id = UUID.randomUUID().toString(),
-                name = stringResource(R.string.cached_playlist)
-            ),
-            songCount = 0,
-            songThumbnails = emptyList(),
-        )
-
```
nyxiereal, line 156: why did you remove the cached playlist? many people use that for well... listening to cached songs

app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryMixScreen.kt
```diff
@@ -368,25 +357,6 @@ fun LibraryMixScreen(
...
-                                Modifier
-                                    .fillMaxWidth()
-                                    .clickable {
-                                        navController.navigate("cache_playlist/cached")
-                                    }
-                                    .animateItem(),
-                            )
-                        }
-                    }
-
```
nyxiereal, line 389: same as other, why? we should have both this AND the cached playlist

kairosci:
@nyxiereal could you take a look at it?  Thank you!

nyxiereal (approved):
LGTM

nyxiereal:
if you want me to merge please reply to confirm

kairosci:
You can merge it

PR #3102, opened 2026-03-01
feat: implement automatic sleep timer with scheduling support

johannesbrauer:
## Problem
Currently, the sleep timer in Metrolist requires manual activation for every session. For users who listen to music at consistent times (e.g., every night at bedtime) is this a bit annoying.

## Cause
Sometimes I forget to activate the sleep timer or I think I forgot it and then I'm going to look at my phone again, so this sometimes prevents me from sleeping.

## Solution
* **UI:** Added `SleepTimerDialog` with Material3 TimePickers. Supports four repeat modes: Daily, Weekdays/Weekends, and Custom.
* **Settings:** Added a new "Sleep Timer" section in Player & Audio settings with a master toggle and schedule configuration.
* **Logic:** * Implemented trigger logic in `PlayerConnection.onPlayWhenReadyChanged` to ensure the timer starts across all local playback sources.
    * Added support for overnight ranges via inverted time checks.
    * Added serialization logic for per-day schedules.
* **Persistence:** Added required DataStore keys to `PreferenceKeys`.

### Technical Notes
- **Serialization:** Per-day schedules (for Weekdays/Weekends and Custom modes) are stored as a mapped string format.
- **Compatibility:** Google Cast is intentionally excluded as the timer targets local playback sessions.
- **Efficiency:** The check only triggers on `paused -> playing` transitions and only if a timer isn't already active.

## Testing
- **Manual Verification:** Verified by triggering playback during scheduled windows across different modes (Daily, Custom, etc.).
- **Scenarios:** Confirmed the timer triggers correctly when starting playback via the App UI, Notification controls, and Bluetooth.
- **Overnight Check:** Verified that ranges crossing midnight (e.g., 22:00 - 02:00) correctly trigger the timer.
- **Tested Devices:** Redmi Turbo 4 pro with android 16 and HyperOS 3 

## Related Issues

- Closes #3101

PR #3090, opened 2026-03-01
feat(playback): add smart sleep timer options and integrate fade-out …

isotjs:
## Problem
The current sleep timer in Metrolist only supports a basic countdown and abrupt stop behavior.
This can interrupt listening experience when users are falling asleep, especially when playback is cut immediately at timer end.

## Cause
`SleepTimer` was designed as a simple delay-based pause mechanism, without:
- an option to wait for the current track to end
- a gradual volume fade before stopping
This limited UX flexibility and made timer shutdown feel harsh.

## Solution
Added a **smart sleep timer** flow in `SleepTimer` with:
  - countdown stop behavior
  - stop at end of current song when timer expires
  - optional fade-out during the final minute
  - Integrated sleep timer fade-out into playback volume calculation in `MusicService` so it works with mute and audio focus ducking.
  - Added new preferences for default timer options:
  - stop-after-current-song
  - fade-out
  - Updated both Player and Queue sleep timer dialogs to expose the two new options.
  - Added user-facing strings in `metrolist_strings.xml` for the new options.

## Testing
Verified timer behavior manually for:
  - normal countdown stop
  - countdown + stop at end of current song
  - countdown + fade-out
  - end-of-song mode + fade-out
  - cancel/clear behavior restoring normal volume
  - Verified option persistence via “Set as default” and dialog reopen checks (Player + Queue).

## Related Issues (edited)
  - Closes #1955
  - Related to #

PR #3142, opened 2026-03-03
(Refactor) sleeptimer UI

johannesbrauer:
## Problem
The new sleep timer features Stop at end of current song and Fade out were very confusing in the Sleep timer dialog, they weren't persistent by default and they had a old ui style. The Automatic sleep timer Dialog also used old check boxes for the days to select.

## Solution
I moved them to the 'PlayerSettings page to the Sleep timer section. I gave them new Material 3 switches and new Material 3 icons. The logic keeps the same by @isotjs .

## Testing
**Tested devices**: Redmi Turbo 4 pro with android 16 and HyperOS 3.

I also tested that these Features still work, but as the logic keeps there shouldn't be any changes.

## Related PR's
- Related to #3102 #3090

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/component/SettingsSleepTimerDialog.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/component/SettingsSleepTimerDialog.kt
@@ -18,7 +18,6 @@ import androidx.compose.ui.draw.scale
 import androidx.compose.ui.res.painterResource
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.OutlinedButton
 import androidx.compose.material3.RadioButton
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
@@ -36,6 +35,23 @@ import androidx.compose.ui.unit.dp
 import com.metrolist.music.R
 import java.time.LocalTime
 import java.time.format.DateTimeFormatter
+import androidx.compose.material3.SingleChoiceSegmentedButtonRow
+import androidx.compose.material3.SegmentedButton
+import androidx.compose.material3.SegmentedButtonDefaults
+import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.ElevatedCard
+import androidx.compose.material3.FilledTonalButton
+import androidx.compose.material3.Button
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.expandVertically
+import androidx.compose.animation.shrinkVertically
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
+import androidx.compose.animation.core.spring
+import androidx.compose.animation.core.Spring
+import androidx.compose.material3.ButtonGroup
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.ExperimentalMaterial3ExpressiveApi
 
 fun decodeDayTimes(raw: String): MutableMap<Int, Pair<String, String>> {
     if (raw.isBlank()) return mutableMapOf()
@@ -61,7 +77,7 @@ private const val DEFAULT_END = "06:00"
 private val WEEKDAY_INDICES = 0..4 // Monday to Friday
 private val WEEKEND_INDICES = 5..6 // Saturday and Sunday
 
-@OptIn(ExperimentalMaterial3Api::class)
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalMaterial3ExpressiveApi::class)
 @Composable
 fun SleepTimerDialog(
     isVisible: Boolean,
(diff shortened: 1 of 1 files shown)


isotjs:
Maybe we should switch the Set as default with a save icon that will be positioned on the left side along the texts on window? It catches my eyes fr fr 😅

johannesbrauer:
> Maybe we should switch the Set as default with a save icon that will be positioned on the left side along the texts on window? It catches my eyes fr fr 😅 

How do you mean that? I actually don't know where you mean, in my opinion this was the best position. But I'm open for other suggestions.

isotjs:
> > Maybe we should switch the Set as default with a save icon that will be positioned on the left side along the texts on window? It catches my eyes fr fr 😅
> 
> How do you mean that? I actually don't know where you mean, in my opinion this was the best position. But I'm open for other suggestions.

I will try commiting it ASAP If I can spare some time :smile:

johannesbrauer:
@isotjs ok, but idk what's the problem with the position now?

isotjs:
> @isotjs ok, but idk what's the problem with the position now?

It being on the middle of the ui gives a newbie vibe we may fix this via either making it like a button style or making it a icon and put it along others after reset.

johannesbrauer:
@isotjs ok, I can make it as a button, because besides the reset button isn't enough space.

isotjs:
> @isotjs ok, I can make it as a button, because besides the reset button isn't enough space.

Yea that would work too, I was going to commit but I couldn't spare time 😞

isotjs (approved):

johannesbrauer:
I have all merge conflicts resolved and tested it again, I also made the 'Set as default' button as a OutlinedButton as @isotjs wanted it. Should be ready to merge.

app/src/main/kotlin/com/metrolist/music/ui/player/Player.kt
```diff
@@ -673,6 +637,17 @@ fun BottomSheetPlayer(
...
+                    OutlinedIconButton(
+                        onClick = {
+                            showSleepTimerDialog = false
+                            playerConnection.service.sleepTimer.start(
+                                minute = -1,
+                            )
+                        },
+                    ) {
+                        Text(stringResource(R.string.end_of_song))
+                    }
```
nyxiereal, line 678: both these buttons should be in the same horizontally aligned row, not in separate columns. there is also no visual feedback for the "set as default" being the current time. my vision is:
- user sets 30 minutes as the default sleep times
- the set as default button turns the user's material you primary color when the sleep timer is at the set default value
- when the user slides to another value, the button changes to its current state (as in the current code). when the user clicks it again it changes to the primary color

app/src/main/res/values/metrolist_strings.xml
```diff
@@ -260,8 +260,10 @@
     <string name="sleep_timer_friday">Friday</string>
     <string name="sleep_timer_saturday">Saturday</string>
     <string name="sleep_timer_sunday">Sunday</string>
-    <string name="sleep_timer_stop_after_current_song">Stop at end of current song when timer ends</string>
-    <string name="sleep_timer_fade_out">Fade out in final minute</string>
+    <string name="sleep_timer_stop_after_current_song_title">Stop at end of current song</string>
+    <string name="sleep_timer_stop_after_current_song_description">Stops the sleep timer at the end of the current playing song</string>
```
nyxiereal, line 264: this probably means that it will stop after the automatic sleep timer in addition to waiting for the current song to end, please specify that clearly

app/src/main/kotlin/com/metrolist/music/ui/screens/settings/PlayerSettings.kt
```diff
@@ -594,6 +604,55 @@ fun PlayerSettings(
                         )
                     )
                 }
```
nyxiereal: make it show the option as disabled, you have to show it instead of hiding it so there is no huge ui shift

app/src/main/kotlin/com/metrolist/music/ui/component/SleepTimerDialog.kt
nyxiereal, line 1: please redesign this dialog using more material 3 expressive components and compact the design slightly

nyxiereal: could you add separators between days for clearer... separation

nyxiereal: the same goes for the weekdays/weekends option, but the custom one is way more severe of a flaw

app/src/main/res/values/metrolist_strings.xml
johannesbrauer: It doens't only mean that it stops after the automatic Sleep timer in addition to the end of the current song, it's meant as global setting to always wait until the end of the current song when the sleep timer is activated, doesn't matter if it's activated from the automatic sleep timer or from the user. So if you activate this the sleep timer will always wait for the current song to end.

app/src/main/kotlin/com/metrolist/music/ui/player/Player.kt
johannesbrauer: Ok I'm changing that, I add it to the PR as soon as possible.

johannesbrauer:
Think everything should be changed as wanted, I added M3E everywhere, hope it looks better now.

## Screenshots:

app/src/main/kotlin/com/metrolist/music/ui/screens/settings/PlayerSettings.kt
johannesbrauer: It's now automatically disabling and enabling with the automatic sleeptimer, cause otherwise this button doesn't make sense. But if you click on it now it opens the Dialog

app/src/main/kotlin/com/metrolist/music/ui/component/SleepTimerDialog.kt
johannesbrauer: Ok I changed this, you can look at the new commits and share if you like the M3E bouncy design or not

johannesbrauer: Changed this.

isotjs:
> Think everything should be changed as wanted, I added M3E everywhere, hope it looks better now.
> 
> ## Screenshots:
> 
> 
> 
> 
> 
> 
> 

Looking waaay more better!

johannesbrauer:
@isotjs btw if you believe the change log now you also made the automatic sleep timer 😂, but idk why they didn´t merged this PR before the new version bump. I extra hurried and worked all night on it so that it´s ready before the new version bump so that the confusing UI don´t come to normal users.

isotjs:
> @isotjs btw if you believe the change log now you also made the automatic sleep timer 😂, but idk why they didn´t merged this PR before the new version bump. I extra hurried and worked all night on it so that it´s ready before the new version bump so that the confusing UI don´t come to normal users.

@nyxiereal fr maybe a hot fix for this would be great

isotjs:
> @isotjs btw if you believe the change log now you also made the automatic sleep timer 😂, but idk why they didn´t merged this PR before the new version bump. I extra hurried and worked all night on it so that it´s ready before the new version bump so that the confusing UI don´t come to normal users.

Ayooo, das habe ich nicht bemerkt 😅

johannesbrauer:
@isotjs also German? I think we should stay at English on GitHub, so that there is everything understandable for everyone, and that newbies don´t get confused. But interesting 😂

isotjs:
> @isotjs also German? I think we should stay at English on GitHub, so that there is everything understandable for everyone, and that newbies don´t get confused. But interesting 😂

No but I learnt it, I just saw ur profile and thought it would be fun to answer in german 😅

johannesbrauer:
@nyxiereal would be nice if you could review it, thanks

PR #3171, opened 2026-03-06
Changes to m3u playlist export

xn-7492:
## Problem
- Authors are seperated by ' - ', but should be seperated by ';'.
- I'm not sure why the video url is not included, because the video id is sufficient to do that.

## Cause
yes

## Solution
- Split multiple authors with ';' instead of ' - '
- Append youtube url with the video id

## Testing
Exported a test playlist in emulator


nyxiereal:
when the workflow completes i will merge

PR #3180, opened 2026-03-07
fix(lastfm): adjust width of login dialog

omardotdev:
## Problem
the lastfm integration's lastfm dialog doesn't really match the width of other dialogs...
IMG_20260307_112725

## Solution
do NOT use fillmaxwidth
IMG_20260307_112709


nyxiereal:
good change!

PR #3217, opened 2026-03-11
feat: remove unused scrollBehavior across pages

omardotdev:
## Problem
most or almost all pages have an unused scrollBehavior parameter guh

## Solution
remove it!!


nyxiereal:
LGTM, but test the screens (scrolling, etc) and then get back to me

omaawr:
@nyxiereal tested and everything seems to work

PR #3240, opened 2026-03-15
fix(cipher): support player.js 74edf1a3 Q-array obfuscated cipher functions

alltechdev:
## Problem
Playback fails for uploaded songs, explicit content, and some regular tracks due to YouTube's player.js (hash 74edf1a3) using Q-array obfuscation that breaks cipher function extraction.

## Cause
YouTube's player.js now uses Q-array obfuscation (`var Q="...".split("}")`) which replaces string literals with array tokens. This breaks regex-based extraction of signature and n-transform function names. The cipher functions also require multiple constant arguments (e.g., `JI(48, 1918, f1(1, 6528, sig))`) instead of the previous single-argument pattern.

## Solution
- Add hardcoded config for player.js hash 74edf1a3 as a fallback when regex extraction fails
- The existing regex-based extraction is preserved and tried first - hardcoded is only used if regex doesn't match
- Support signature chain: `JI(48, 1918, f1(1, 6528, sig))` with preprocessing
- Support n-transform: `GU(6, 6010, n)` with multiple constant arguments
- Generate wrapper functions for multi-argument cipher functions
- Detect wrapper vs direct functions via `func.length === 1` in JavaScript

## Testing
- Tested regular song playback
- Tested uploaded/private song playback (MUSIC_VIDEO_TYPE_PRIVATELY_OWNED_TRACK)
- Tested explicit content playback
- Tested age-restricted content (via TVHTML5 fallback)
- Verified n-transform applied correctly
- Verified signature deobfuscation works

To view cipher logs:
```
adb logcat | grep -E "(Metrolist_Cipher|YTPlayerUtils|N-TRANSFORM|HARDCODED)"
```

## Related Issues
- Related to cipher/playback issues with recent YouTube changes

## Release Notes

* **Improvements**
  * Enhanced playback reliability with improved error handling and diagnostic logging for music stream processing.
  * Added automatic cache management for YouTube player data, optimizing app performance and reducing redundant downloads.
  * Better error detection and recovery when handling protected content and age-restricted tracks.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
@@ -80,14 +80,18 @@ object YTPlayerUtils {
         audioQuality: AudioQuality,
         connectivityManager: ConnectivityManager,
     ): Result<PlaybackData> = runCatching {
-        Timber.tag(logTag).d("Fetching player response for videoId: $videoId, playlistId: $playlistId")
-        // Debug: Log ALL playback attempts
-        println("[PLAYBACK_DEBUG] playerResponseForPlayback called: videoId=$videoId, playlistId=$playlistId")
+        Timber.tag(TAG).d("=== PLAYER RESPONSE FOR PLAYBACK ===")
+        Timber.tag(TAG).d("videoId: $videoId")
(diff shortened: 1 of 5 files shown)


alltechdev:
Responding to coderabbit review. 

All flagged issues are pre-existing in main branch - not introduced by this PR. These are out of scope for this fix.

PR #3274, opened 2026-03-18
RFC: Enhanced library navigation - Option 2: Tabbed library page

### LINKED ISSUE #3267: RFC (PR proposal): Enhanced library navigation [Feedback wanted]
## Checklist

## Feature description

I am willing to work on improving the library navigation and would like to discuss different options before going on with the implementation.

### Observations

Navigating between different library categories, eg. songs, artists, albums and so on, is something some users might do fairly frequently, and this is made needlessly annoying because of the following issues.

I would like to improve this by revamping the library page, and/or allowing users to configure pinned categories directly in the navigation bar in the settings.

### Current problems

- The tag filters are a small touch target, making it hard to reach
- Navigating between two of these pages requires two taps, first tapping the first filter acting as a "back" button, then the desired filter
- While a minor problem, this two-level filter system might be unintuitive to new users, with the tags changing depending on the current selection, and some tags acting as more of a navigation system than actual filters.

### Current state

Two-level filter system, with first level acting as navigation, and second level comprising of a "back" (un)filter followed by actual filters. Small touch targets are hard to reach and require two taps for navigation, and the home (unfiltered) page feels like a hap-hasard selection of items, with no clear explanation of what is shown there and why.

### Option 1: Expandable carousels

| Illustration | Description |
| :---: | ----- |
|  | Main library page is split in one expandable carousel for each of the library category, similar to the sections on the homepage or the results on the search page.<br/><br/>While this does not address the "two taps to navigate between categories" problem, it provides a nicer overview of the library's content in each category, and is consistent with other parts of the application. |

### Option 2: Tabbed library page

| Illustration | Description |
| :---: | ----- |
|  | While their use has mostly faded away with Material 3, tabbed pages [are indeed a valid M3E component](https://m3.material.io/components/tabs/overview). This allows switching between categories with just one tap, or swiping can be used if not enabled for individual item actions.<br/><br/>This also leaves space for the actual filtering tags, giving them a more consistent meaning. |

### Option 3: Navigation bar & overflow shelve

| Illustration | Description |
| :---: | ----- |
| <br/>↓<br/>Tap on last tab<br/>↓<br/> | This option would see the library page as we know it completely deleted in favor of a wikipedia-style "overflow shelve" allowing one to access the various library pages through a bottom shelve shown upon tapping the last navigation tab like on wikipedia.<br/><br/>Of course, this would work best in combination with the ability to pin certain library pages to the navigation bar. |

### Optionally: Allow pinning library pages

| Illustration | Description |
| :---: | ----- |
|  | Without removing the library page, allow users to pin any category as a main page on the navigation bar for faster navigation between categories.<br/><br/>In addition to being editable in the settings, these pins could be configured in a more discoverable way, such as a "pin" button displayed where appropriate depending on the option selected above.

## Additional information

It goes without saying I do not intend or expect for all of these options to be available and maintained in the long run, however I think revamping the library page along with allowing users to pin select categories in the navigation bar would greatly improve the experience of the app.

In my opinion, option 1 would work best in addition to the ability to pin categories to the navigation bar, while option 2 could work very well no matter the case.

I'd love to get some feedback from users and maintainers alike so we can choose the best option before I try my hand at implementing the proposed changes !

Cheers,
Lurux.

Lurux:
This PR implements option 2 (Tabbed library page) of RFC #3267 (Enhanced library navigation).

This allows users to switch between categories with a single tap, and makes filters more intuitive, giving them a consistent behavior across all pages of the library.

As per user feedback, commit cfa0b05cf2ea0def6c17740a271cb3366f080a23 adds icons to the library tabs, although it is my opinion that adding icons to these library tabs makes the interface feel more crowded. Given the current filter chips don't feature such icons, the merits of including them here should be discussed. Of course, the presence of these icons could be made configurable, but these are hard-coded as of this first draft.

| With icons | Without icons |
| :---: | :---: |
|  |  |

I have not implemented swiping between tabs as of yet, but this could be made configurable when swiping individual items/songs is disabled.

I have not implemented a "home" / mix / overview library tab as of yet, but such a tab could be re-introduced if we find something meaningful to house there, such as featuring the speed dial from the homepage or a few "top" songs (most listened) from the past week.

---

Resolves #3267

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryScreen.kt
@@ -12,7 +12,6 @@ import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.statusBars
 import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.PrimaryScrollableTabRow
 import androidx.compose.material3.Tab
@@ -25,7 +24,6 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.platform.LocalDensity
-import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.zIndex
@@ -57,32 +55,27 @@ fun LibraryScreen(navController: NavController, scrollBehavior: TopAppBarScrollB
             Tab(
                 selected = true,
                 onClick = { filterType = LibraryFilter.SONGS },
-                text = { Text(text = stringResource(R.string.filter_songs)) },
-                icon = { Icon(painter = painterResource(R.drawable.music_note), contentDescription = stringResource(R.string.filter_songs)) }
+                text = { Text(text = stringResource(R.string.filter_songs)) }
             )
             Tab(
                 selected = false,
                 onClick = { filterType = LibraryFilter.ARTISTS },
-                text = { Text(text = stringResource(R.string.filter_artists)) },
-                icon = { Icon(painter = painterResource(R.drawable.person), contentDescription = stringResource(R.string.filter_artists)) }
+                text = { Text(text = stringResource(R.string.filter_artists)) }
             )
             Tab(
                 selected = false,
                 onClick = { filterType = LibraryFilter.ALBUMS },
-                text = { Text(text = stringResource(R.string.filter_albums)) },
-                icon = { Icon(painter = painterResource(R.drawable.album), contentDescription = stringResource(R.string.filter_albums)) }
+                text = { Text(text = stringResource(R.string.filter_albums)) }
             )
             Tab(
                 selected = false,
                 onClick = { filterType = LibraryFilter.PLAYLISTS },
-                text = { Text(text = stringResource(R.string.filter_playlists)) },
-                icon = { Icon(painter = painterResource(R.drawable.library_music_outlined), contentDescription = stringResource(R.string.filter_playlists)) }
+                text = { Text(text = stringResource(R.string.filter_playlists)) }
             )
             Tab(
                 selected = false,
                 onClick = { filterType = LibraryFilter.PODCASTS },
-                text = { Text(text = stringResource(R.string.filter_podcasts)) },
-                icon = { Icon(painter = painterResource(R.drawable.radio), contentDescription = stringResource(R.string.filter_podcasts)) }
+                text = { Text(text = stringResource(R.string.filter_podcasts)) }
             )
         }
 


app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryScreen.kt
```diff
@@ -43,26 +60,68 @@ fun LibraryScreen(navController: NavController) {
...
-            LibraryFilter.PLAYLISTS -> LibraryPlaylistsScreen(navController, filterContent)
-            LibraryFilter.SONGS -> LibrarySongsScreen(
-                navController,
-                { filterType = LibraryFilter.LIBRARY },
+    Column {
+        PrimaryScrollableTabRow(
+            selectedTabIndex = filterType.ordinal,
+            modifier = Modifier
+                .graphicsLayer {
+                    translationY = scrollBehavior.state.heightOffset + TopAppBarExpandedHeight.toPx() + statusBarHeight
```
Lurux, line 49: I'm a bit iffy about what I did to properly position the library tabs here, however I didn't find a better way to do it. Current behavior is the tabs "stick" to the top of the screen when scrolling down, but this should be easy enough to modify.

isotjs: What about M3's native scroll-sticky?

Lurux: I'll take a look ! Problem I had here is the tab bar is outside the LazyColumn because I didn't want to duplicate it, but for some reason it doesn't take the status bar and app bar in consideration like the LazyColumn does when positioning.

Depending on the intended effect we might revert to passing it to each screen like we did with the filters tags, but that feels kinda crappy and would probably (?) break the animations

Lurux: I did try putting it in the stickyHeader inside the LazyColumn but I seem to remember it still didn't position correctly

isotjs: > I did try putting it in the stickyHeader inside the LazyColumn but I seem to remember it still didn't position correctly

I will check this tomorrow maybe I can figure out a better way for this

Lurux: I'll check back on that but I seem to remember we only need to adjust for the statusbar height when putting it inside the LazyColumn

isotjs: The `graphicsLayer { translationY }` approach only applies a visual offset. It doesn't actually move the layout space, which means during AppBar collapse/expand animations the TabRow will either overlap the content or leave a gap behind.

I think we can solve this cleanly by going back to the upstream `filterContent` lambda pattern: pass the `TabRow` as a composable lambda into each sub-screen and place it as a `stickyHeader` inside their respective `LazyColumn`/`LazyGrid`. That way the AppBar offset accounting becomes completely unnecessary since the column's `contentPadding` insets already handle that.

For `LazyGrid` screens, the sticky header item just needs `span = { GridItemSpan(maxLineSpan) }` like the old filter items had.

I'll try to take a look at this tomorrow again and see if I can put together a working version.

Lurux: I checked that the current version doesn't have any visual glitches like those you mentioned ofc, however I reckon that solution might be cleaner (although I hate the idea of needlessly duplicating that, even if it's just some passing around)

Lurux: I'm not sure swipe between tabs and indicator animations would work properly if the tab bar is recomposed along with the actual contents every tile we switch tab ? I'm not an expert there, will take a look later on

isotjs: The lambda passing is not really duplication, the TabRow itself still lives in one place, we are just passing a reference down like the old filterContent pattern did.

On the animation concern, since swipe is not implemented yet, there is no recomposition issue to worry about right now. When swipe does get added, it will need a HorizontalPager anyway and the architecture will have to change regardless of which positioning approach we go with. So I would not let that block the cleaner solution for now.

Lurux: Ok, I see, thanks for the insight. Will tinker with that after my current work 😄

Lurux:
See also https://github.com/MetrolistGroup/Metrolist/issues/3267#issuecomment-4084406476 and the following conversation for styling discussion

app/src/main/kotlin/com/metrolist/music/ui/screens/library/LibraryScreen.kt
isotjs: Good luck with it mate 🙂

PR #3289, opened 2026-03-19
feat: use low contrast on account settings popup

omardotdev:
## Problem
the slightly high contrast of the components the account settings dialog doesn't really look good with the bright background

## Solution
make it darker!!

## Summary by CodeRabbit

* **New Features**
  * Added low-contrast rendering option for the Account Settings interface, improving visual clarity and accessibility for users who prefer reduced contrast displays.


johannesbrauer:
Thanks for fixing 👍

PR #3296, opened 2026-03-20
Improve search focus and navigation behavior

### LINKED ISSUE #3077: Make tapping the search icon bring up the keyboard
### Checklist

### Feature description

When tapping the search icon in the navigation bar, it should bring up the keyboard

So I can search with one hand 

### Additional information

_No response_

saivijaychandan:
## Problem
Tapping the search navigation icon while already on the search screen or search results screen did not bring up the keyboard.

## Cause
- The `scrollToTop` signal used a boolean, so repeated taps on the same screen were silently ignored since the value never changed.
- `currentBackStackEntry` was captured inside a `remember` lambda, causing it to go stale and return null when reading the `SavedStateHandle`.
- The search results screen route (`search/{query}`) was not recognized as part of the search nav item's selected state in `isRouteSelected`.
- The `DisposableEffect` lifecycle observer was racing against the keyboard show logic, calling `hide()` after `show()`.
- In `AppNavigationBar`, `isSelected` was captured at `LaunchedEffect` creation time and never updated inside the coroutine.

## Solution
- Replaced the boolean `scrollToTop` flag with an incrementing integer counter `scrollToTopCount` so repeated taps always trigger the effect.
- Passed `backStackEntry.savedStateHandle` directly into `SearchScreen` and `OnlineSearchResult` instead of looking up the back stack during composition.
- Added `isHandlingScrollToTop` guard to prevent the lifecycle observer from hiding the keyboard during the show sequence.
- Updated `isRouteSelected` to match `search/{query}` route template to the search nav item.
- Used `rememberUpdatedState` for `isSelected` in `AppNavigationBar` and `AppNavigationRail` so the coroutine always reads the latest value.
- Used `navController.currentBackStackEntry` live inside the `onNavItemClick` lambda instead of a captured reference.

## Testing
- Tapping search icon while on search screen opens keyboard.
- Tapping search icon repeatedly works every time.
- Tapping search icon while on search results screen focuses the search field and shows suggestions overlay.
- Tapping search icon after clearing query letter by letter works correctly.
- Keyboard is hidden when the music player is expanded.

## Related Issues
- Closes #3077 
- Related to #3077 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search input focus behavior and keyboard visibility when navigating between search screens.
  * Fixed search tab selection handling for repeated interactions.

* **Refactor**
  * Restructured navigation state management to better handle search-related screen transitions.
  * Enhanced focus and keyboard management during search operations.


app/src/main/kotlin/com/metrolist/music/ui/component/AppNavigation.kt
```diff
@@ -1,211 +1,221 @@
...
+    }
+
+    @Composable
+    fun AppNavigationRail(
+        navigationItems: List<Screens>,
+        currentRoute: String?,
+        onItemClick: (Screens, Boolean) -> Unit,
+        modifier: Modifier = Modifier,
+        pureBlack: Boolean = false,
+        onSearchLongClick: (() -> Unit)? = null
```
nyxiereal, line 64: why the hell did you just add 4 spaces before each line here???? wtf??? revert that

saivijaychandan: Sorry, didn't check that

PR #3302, opened 2026-03-20
feat: add song countdown for sleep timer, redesign sleep timer ui

Ryfters:
## Changes
 - Add song countdown
 - Redesign UI (See screenshots, For End of Song, make a 1 song countdown)
 - Defaults will also take in account the countdown type (song, time finish, time)
 
Screenshot_20260320-190242
Screenshot_20260320-190245

## Testing
Tested all new options, with and without crossfade

Known bug: If a user manually goes to the next song when only 1 song remains, it will end the timer without pausing. 

### For maintainers
 - The fade out part is not implemented yet. It seems to be disabled in this build. If it was for the previous UI redesign, I can now fit a fade out check button where the End of Song button was.
 - Identical sleep timer UIs (one in app/src/main/kotlin/com/metrolist/music/ui/player/Player.kt, one in app/src/main/kotlin/com/metrolist/music/ui/player/Queue.kt). I disabled the one in player as it didn't affect anything, I can update it like the new version if required.

## Summary by CodeRabbit

* **New Features**
  * Sleep timer now supports three modes: minutes, finish-current-song, and song-count.
  * You can set and persist a preferred sleep-timer mode and value as the default.

* **Refactor**
  * Redesigned sleep-timer dialog with a unified selection UI for mode and value.

* **Behavior Changes**
  * Crossfade will only start when the sleep timer has one remaining song.
  * Fade-out/volume-fade option removed.

* **Localization**
  * Improved sleep-timer strings and plurals for minutes and songs.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
+import timber.log.Timber
 import kotlin.time.Duration.Companion.minutes
 
 class SleepTimer(
@@ -25,7 +26,6 @@ class SleepTimer(
 ) : Player.Listener {
     companion object {
         private const val TIMER_TICK_MS = 1000L
-        private const val FADE_OUT_WINDOW_MS = 60_000L
 
         const val TIME = 0
         const val TIME_FINISH = 1
@@ -40,6 +40,7 @@ class SleepTimer(
         private set
     var type by mutableIntStateOf(TIME)
         private set
+    // fadeOut is disabled for now
     var fadeOutEnabled by mutableStateOf(false)
         private set
     val isActive: Boolean
@@ -47,14 +48,12 @@ class SleepTimer(
 
     fun start(
         value: Int,
-        _type: Int,
-        fadeOut: Boolean = false,
+        timerType: Int,
     ) {
         sleepTimerJob?.cancel()
         sleepTimerJob = null
-//        updateVolumeMultiplier(1f)
-        fadeOutEnabled = fadeOut
-        type = _type
+        updateVolumeMultiplier(1f)
+        type = timerType
 
         if (type == SONGS) {
             triggerTime = -1L
@@ -102,7 +101,7 @@ class SleepTimer(
         fadeOutEnabled = false
         remainingSongs = 0
         triggerTime = -1L
-//        updateVolumeMultiplier(1f)
+        updateVolumeMultiplier(1f)
     }
 
     override fun onMediaItemTransition(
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
```diff
@@ -111,70 +89,71 @@ class SleepTimer(
...
     override fun onMediaItemTransition(
         mediaItem: MediaItem?,
         reason: Int,
     ) {
-        if (pauseWhenSongEnd) {
-            completeTimerAndPause()
+        if (remainingSongs > 0) {
+            if (--remainingSongs == 0){
+                completeTimerAndPause()
+            }
```
isotjs, line 114: notifySongTransition() already does the exact same decrement. If MusicService calls notifySongTransition() on song change and this listener might also fire, remainingSongs drops twice per transition. This is likely the root cause of the "manually skipping ends the timer early" bug you mentioned.

app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
```diff
@@ -111,70 +89,71 @@ class SleepTimer(
...
-            completeTimerAndPause()
-        }
-    }
+//    override fun onPlaybackStateChanged(
+//        @Player.State playbackState: Int,
+//    ) {
+//        if (playbackState == Player.STATE_ENDED && pauseWhenSongEnd) {
+//            completeTimerAndPause()
+//        }
+//    }
```
isotjs: This needs to come back, adapted for remainingSongs. When the playlist ends and STATE_ENDED fires, the timer is never cleared now.

app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
```diff
@@ -23,85 +23,63 @@ class SleepTimer(
...
-        stopAfterCurrentSong: Boolean,
-        fadeOut: Boolean,
+        value: Int,
+        _type: Int,
+        fadeOut: Boolean = false,
     ) {
         sleepTimerJob?.cancel()
         sleepTimerJob = null
-        updateVolumeMultiplier(1f)
+//        updateVolumeMultiplier(1f)
```
isotjs: Commenting **these** out means if a previous session left the volume multiplier below 1.0 mid-fade, it never resets. These should stay active unconditionally, or at minimum be guarded with if (fadeOutEnabled) rather than removed entirely.

app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
```diff
@@ -111,70 +89,71 @@ class SleepTimer(
...
+//    }
+
+//    private fun volumeMultiplierForRemainingTime(remainingMs: Long): Float {
+//        if (remainingMs >= FADE_OUT_WINDOW_MS) return 1f
+//        return (remainingMs.toFloat() / FADE_OUT_WINDOW_MS).coerceIn(0f, 1f)
+//    }
+//
+//    private fun updateVolumeMultiplier(multiplier: Float) {
+//        onVolumeMultiplierChanged(multiplier)
+//    }
```
isotjs: I understand fade-out is deferred, but commenting out the implementation while keeping fadeOutEnabled as a live field and FADE_OUT_WINDOW_MS as a constant leaves the class in an inconsistent state. Either keep the functions (even if not called yet) or mark the feature clearly as a stub.

app/src/main/kotlin/com/metrolist/music/ui/player/Player.kt
```diff
@@ -572,114 +546,169 @@ fun BottomSheetPlayer(
...
+////                                        )
+////                                    },
+////                                ) {
+////                                    Text(stringResource(R.string.end_of_song))
+////                                }
+//                    }
+//                }
+//            },
+//        )
+//    }
```
isotjs: Both Player.kt and Queue.kt entry points should reflect the new state model consistently. Disabling one is a UX regression for users who open the timer from the player screen directly.

isotjs:
@coderabbitai review

nyxiereal:
DO NOT TOUCH strings.xml!!!!!!!!!

adrielGGmotion:
I think they touched strings.xml... 🧐

Ryfters:
oops... 🫣

Ryfters:
@coderabbitai review

app/src/main/kotlin/com/metrolist/music/playback/SleepTimer.kt
Ryfters: notifySongTransition() is only called on the crossfade transition, as this doesn't trigger onMediaItemTransition(). I meant to say that manually skipping correctly decrements the song counter, but when there are no more songs, it deactivates the sleep timer, but without pausing the song like it should.

isotjs:
Redesign the buttons to match the apps standards

nyxiereal:
Stale + unwanted kinda

PR #3330, opened 2026-03-23
fix(Lyrics.kt): simplify if expression

omardotdev:
## Problem
you dont need to repeat `Text` on multiple if statements when the only difference is 2 parameters guh

## Solution
make it once!!

## Summary by CodeRabbit

* **Refactor**
  * Consolidated lyrics display styling logic to simplify rendering paths; improves maintainability while preserving the visual appearance and behavior of lyrics (no user-facing changes).


omaawr:
@coderabbitai ignore

PR #3332, opened 2026-03-23
fix(Player): awkward padding on buttons when lyrics is hidden

omardotdev:
## Problem
theres a giant padding in the middle of the buttons right next to the lyrics when you tap on the lyrics button..

IMG_20260323_193844

## Solution
fix it!!
IMG_20260323_202123

@coderabbitai ignore


nyxiereal:
not needed now

johannesbrauer:
@nyxiereal why it isn't needed, I was also on fixing it

johannesbrauer:
LGTM to me, I tried it the same way.

PR #3361, opened 2026-03-27
Add Netease Cloud Music lyrics provider with EAPI integration

grill-glitch:
> I myself cannot fix bugs,but robots can :)

## Release Notes

* **New Features**
  * Added NetEase Cloud Music as a new lyrics provider option for enhanced lyrics sourcing and discovery
  * Introduced a configurable settings toggle to enable or disable NetEase Cloud Music lyrics fetching
  * Implemented support for translated lyrics, allowing users to view both original and translated content together in the lyrics view


adrielGGmotion:
We appreciate you fixing the CodeRabbit issues but after thinking about it we've decided not to add NetEase as a provider. LRCLib and Kugou already cover the same ground, and maintaining EAPI encryption against an external service long-term isn't something we want to take on. Closing this, and we won't be revisiting it.

Issue #3367, opened 2026-03-28
BUG: App not compatible with Advanced EQ apps requiring audio capture


okibcn:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

1. install an advanced audio processing EQ such as Rootless JamesDSP from [Google Play](https://play.google.com/store/apps/details?id=me.timschneeberger.rootlessjamesdsp) or its [GitHub repo](https://github.com/timschneeb/RootlessJamesDSP).
2. Open ArchiveTune (bug last verified with latest nightly on 2026.03.23)
3. Play any track
4. Activate JamesDSP
5. ERROR: Audio is silent

### Expected behavior

The expected behavior is that the audio should go through the advanced EQ and play through speakers or headphones or any other bluetooth audio device. Instead ArchiveTune is silent.

Advanced EQ tested with similar Apps such as OpenTune, SimpMusic, and ArchiveTune and the problem is happening only with Metrolist.

### Actual behavior

When Advanced EQ is activated, the audio output of Metrolist is disbled.

### Logs

logcat not needed. The problem is related to the App permissions.

### Metrolist version

Nightly on 2026-03-27

### Android version

Android 14

nyxiereal:
Duplicate of #1730.

PR #3368, opened 2026-03-28
Allow Audio Playback Capture

okibcn:
## Problem
Metrolist can't be used along with an advanced EQ such as JamesDSP or other advanced DSP that requires audio capture enabled.

## Cause
The app’s audio cannot be recorded because Android is enforcing playback-capture protection by default unless the app explicitly allows it. The cause is the problem is the missing setting in AndroidManifest.xml

## Solution
Added 'android:allowAudioPlaybackCapture="true"' to the '<application>' tag in 'AndroidManifest.xml'.

```xml
<application
    android:name=".App"
    android:allowBackup="true"
    android:allowAudioPlaybackCapture="true"
    ... >
```
## Testing
The code was compiled and tested with rootless JamesDSP advanced DSP EQ.

## Related Issues
- Related to #3367

## Summary by CodeRabbit

* **New Features**
  * Enabled audio playback capture for the app and its media playback, allowing permitted system components and features to capture the app's audio output. No permissions, activities, services, UI, or user-facing settings were added or removed; overall app behavior and controls remain unchanged.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -1067,7 +1067,11 @@ class MusicService :
                 android.media.AudioAttributes.Builder()
                     .setUsage(android.media.AudioAttributes.USAGE_MEDIA)
                     .setContentType(android.media.AudioAttributes.CONTENT_TYPE_MUSIC)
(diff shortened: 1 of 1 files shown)


okibcn:
The added line in AndroidManifest.xml was not enough. The PR requires more investigation.

PR #3436, opened 2026-04-03
fix(sync): preserve downloaded songs in playlist during sync

### LINKED ISSUE #3309: Downloaded songs under regional restrictions disappear from the playlists
### Checklist - Required

1. find a song that has regional restrictions (e.g., a Japanese song available only in Japan).
2. fly to a country without those restrictions (or just turn on a VPN) and add this song to your Liked or any other playlist.
3. download the song.
4. go back home (or turn off the VPN).
5. see that your song has vanished from your playlist.

### Expected behavior

I expect the song to stay in my playlists.

### Actual behavior

Song is gone.

_No response_

### Logs

Don't think we need logs for this.

### Metrolist version

13.3.0

### Android version

Android 11

### Additional information

I wasn't sure whether to label this as a bug or an enhancement. The main problem is YouTube's regional policy; I assume that when you fetch a playlist in a restricted region, those songs don't even appear in the list. I'm not sure if there’s a way to distinguish between deleted songs and those that are hidden due to region changes, but it would be very convenient if this worked.

kairosci:
## Problem
Downloaded songs under regional restrictions disappear from playlists after syncing. When a user downloads a song in a region where it's available and then returns to a restricted region, the playlist sync removes the song because YouTube no longer returns it in the playlist.

## Cause
The `syncPlaylist` function in SyncUtils completely clears and rebuilds the playlist based on the remote songs. When songs are not returned by YouTube due to regional restrictions, they are removed from the local playlist during sync, regardless of whether the user has downloaded them locally.

## Solution
Modified `syncPlaylist` to preserve downloaded songs that are no longer in the remote playlist:
1. Before clearing the playlist, collect IDs of songs that are marked as downloaded
2. After syncing remote songs, re-add any previously downloaded songs that are no longer in the remote playlist
3. This ensures locally downloaded songs remain in the playlist even when they're not available remotely due to regional restrictions

## Testing
The fix preserves downloaded songs during playlist sync. Songs that have been downloaded locally will remain in the playlist even when YouTube doesn't return them due to regional restrictions.

## Related Issues
- Closes #3309

## Summary by CodeRabbit

* **Bug Fixes**
  * Playlist sync now preserves locally downloaded tracks that are no longer present remotely, appending them to the end of the playlist instead of removing them during sync.

* **New Features**
  * Songs now expose an offline/downloaded state so the app can reliably indicate which tracks are available for offline playback.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
@@ -1481,7 +1481,7 @@ class SyncUtils @Inject constructor(
 
                     val localSongsBeforeSync = database.playlistSongs(playlistId).first()
                     val downloadedSongIds = localSongsBeforeSync
-                        .filter { it.song.isDownloaded || it.song.dateDownload != null }
+                        .filter { it.song.song.isDownloaded || it.song.song.dateDownload != null }
                         .map { it.song.id }
                         .toSet()
 


kairosci:
@mostafaalagamy PTAL, thank you!

nyxiereal:
you dont have to ping mo for every PR, you can ping me, i usually review stuff within 24h, sometimes even 12h

kairosci:
Ok, sorry!

nyxiereal:
test it that this wont cause issues when you:
1. download multiple songs on latest release
2. upgrade to your build
3. check if it crashes (if yes, you have to add a database migration)

kairosci:
It doesn't seem to crash or have any issues on my device, so I don't think a migration is necessary

nyxiereal:
okie dokie

ssslakter:
@kairosci Thanks for fixing the issue! But I was thinking: wouldn't it be a problem if a user deleted a song from their YouTube playlist, but it still showed up in the app? Or is that handled in a different way?

Issue #3351, opened 2026-03-26
Audio clarity issue & autoplay not working


Vaishnav-188:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

Open the application
Play any song
Compare the audio quality with other music apps

### Expected behavior

Music should play in high quality (similar to standard streaming apps with clear and crisp audio).

### Actual behavior

Audio playback has reduced clarity and sounds like low bitrate streaming.

### Logs

No major crashes observed during playback.
Player initializes successfully and audio starts without errors.
However, audio output quality is low despite stable playback.
No explicit error messages related to audio decoding or streaming found in logs.

(Optional – if you can attach Logcat)

Please refer to attached Logcat output during playback for more details.

### Metrolist version

13.1.1

### Android version

14

### Additional information

The application works smoothly, but the audio quality of the music playback is noticeably lower compared to other music platforms.
The sound feels compressed and lacks clarity, especially when using headphones or high-quality speakers.

PR #3441, opened 2026-04-04
[wontfix] feat: remove low quality audio option

kairosci:
## Problem
The audio quality setting included a "Low" option, which is a commercial tactic used by streaming services to allow non-paying users to have a basic listening experience. As an open-source non-commercial application, this option has no legitimate use case.

## Cause
The AudioQuality enum originally included LOW as one of the available options, inherited from commercial streaming service patterns.

## Solution
- Removed the LOW audio quality option from the AudioQuality enum
- Updated the format selection logic in YTPlayerUtils to remove LOW case
- Removed LOW from the UI settings in PlayerSettings
- Removed corresponding string resource

## Testing
Build completed successfully.

## Related Issues
- Related to #3351

## Summary by CodeRabbit

* **Changes**
  * Removed the LOW audio quality option from playback settings. Users now have access to AUTO and HIGH quality options only.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/constants/PreferenceKeys.kt
+++ b/app/src/main/kotlin/com/metrolist/music/constants/PreferenceKeys.kt
@@ -111,7 +111,6 @@ val AudioQualityKey = stringPreferencesKey("audioQuality")
 enum class AudioQuality {
     AUTO,
     HIGH,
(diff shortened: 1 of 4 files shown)


nyxiereal:
nah, this option should be kept for people who want to save internet data

PR #3502, opened 2026-04-09
perf(lyrics): implement concurrent fetching, fix lyrics fetch ordering, and optimize LyricsPlus server selection

ibratabian17:
## Problem
Lyrics loading times are excessively slow, and the app was ignoring the user's configured lyrics provider priority, potentially returning lower-priority lyrics. The app previously fetched lyrics from providers sequentially, meaning a slow response or timeout from a single provider would bottleneck the entire loading process. Additionally, the `LyricsPlusProvider` suffered from latency because it tested its list of fallback servers one by one on every single request without remembering which one actually worked.

## Cause
In `LyricsHelper.kt`, the `getLyrics` function used the default `lyricsProviders` list instead of the user's custom `providerOrder` preferences, completely bypassing the priority system. Furthermore, it used a sequential `for` loop to iterate through the providers, making the total wait time cumulative. `LyricsPlusProvider.kt` iterated through `baseUrls` blindly from top to bottom every time. 

## Solution
- **Fixed the ordering bug by ensuring `getLyrics` uses `orderedProviders` based on user preferences rather than the default provider list.**
- Refactored `getLyrics` in `LyricsHelper.kt` to fetch from all enabled providers concurrently using Kotlin Coroutines and a `Channel`.
- Implemented a priority indexing system (`bestIndex`) that collects concurrent results but ensures the highest-priority provider always wins. It safely holds onto faster results from lower-priority providers until it can guarantee no higher-priority provider will succeed.
- Updated `getAllLyrics` to also fetch concurrently and safely append results to a synchronized list.
- Added a `lastWorkingServer` variable in `LyricsPlusProvider.kt` to cache and prioritize the most recently successful server on subsequent fetches.
- Fixed the `duration` parameter format (removed `/ 1000`) and explicitly specified `source` parameters in `LyricsPlusProvider`.

## Testing
- Tested fetching lyrics for various songs to confirm significantly faster load times.
- Verified that the concurrent channel correctly respects the user's configured provider priority order (prevents a fast, low-priority provider from overwriting a slower, high-priority provider).
- Confirmed `LyricsPlusProvider` successfully caches the working server and prioritizes it on the next song. 
- Simulated network delays and timeouts to ensure the Coroutine jobs handle cancellations properly.

## Related Issues
- Closes #
- Related to #

## Summary by CodeRabbit

* **New Features**
  * Added LyricsPlus as a new lyrics provider option

* **Improvements**
  * Faster lyrics retrieval via concurrent, tiered multi-provider coordination and smarter winner selection
  * LyricsPlus reliability improved by prioritizing recently successful servers
  * More accurate provider requests and safer handling when the network is unavailable


nyxiereal (approved):

nyxiereal:
check out the two coderabbitai comments

ibratabian17:
not the coderabbit keep asking refactor for 23987298178937x times

nyxiereal (approved):

PR #3510, opened 2026-04-10
Fix for Artworks on Android Auto

FireLion137:
## Problem
On Android Auto, as of my understanding, artworkUris are cached automatically by the AA system and on a different way, therefore it should be useless to try to use cached images, so the cached system was not working for the same reason and the songs with artworkUri found in cache would just appear without artwork.

## Cause
I am not sure of the real cause of this problem, but by searching online it seems that AA does not have real permission to access the files obtained by the Uri on the device.

## Solution
- A simple solution would have been removing the case where it tries to use cached image on the phone and always use url.toUri, because it works even in offline mode as there is indeed automatic caching on AA, BUT if the songs are loaded while offline, AA would cache an empty artwork and would not refresh it unless the Uri of that song changed (making the caching useless).
- So the solution here just uses artworkData instead of artworkUri, with an automatic download and caching.

## Testing
- Built successfully, tested on my device and DHU.
- It works with cached images retreived by using the app on the phone.
- If the songs artworks are not downloaded yet, the user just need to exit and enter again to see them, because the download happens as soon the user enters the first time.

## Summary by CodeRabbit

* **Chores**
  * Updated album artwork handling in media metadata for media playback.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MediaLibrarySessionCallback.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MediaLibrarySessionCallback.kt
@@ -765,12 +765,14 @@ constructor(
         ).build()
 
     private fun Song.toMediaItem(path: String, isPlayable: Boolean = true, isBrowsable: Boolean = false): MediaItem {
-        val artworkUri = song.thumbnailUrl?.let {
-            val snapshot = context.imageLoader.diskCache?.openSnapshot(it)
-            if (snapshot != null) {
-                snapshot.use { snapshot -> snapshot.data.toFile().toUri() }
-            } else {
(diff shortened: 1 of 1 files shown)


FireLion137:
As the bot said, the fix that he used works, but it is required to make Song.toMediaItem suspend and for larger playlists the first browse would take a lot of time. Maybe it is preferred to be the way i proposed, with the artworks downloading in the background while you can use, even without the artworks, the playlist.
But this is a problem just for the first time (or two times for over about 200 songs) the playlist is browsed, so your choice, this does not even matter if you just browser the playlist on the phone

nyxiereal:
@iziosettanta what are you doing

PR #3518, opened 2026-04-10
Add missing support for lyrics on Android Auto

FireLion137:
## Problem
There isn't a viable way to access Lyrics on Android Auto.

## Cause
Probably the absence of the Lyrics is caused by the actual limitations of the system itself.

## Solution
Should not be a "feature" by itself but rather adding a missing part already present on the main app.
Adding missing support for lyrics on Android Auto by making them a browsable section in the root of AA.
Probably this is not the best way to enjoy the lyrics visually, but i think this is the simplest way to access them on AA.
- Add lyrics as a browsable section in Android Auto
- Real-time lyrics synchronization and romanization if present
- Fetch and cache lyrics from providers if not present in the local database

## Testing
- Built successfully, tested on my device and DHU.
- The only problem there is with this implementation is that the lyrics themself are clickable and opens an empty list. I did not find a way to make them unclickable or look like ignoring clicks.

## Summary by CodeRabbit

* **New Features**
  * Browse lyrics from a dedicated Lyrics root while a track is playing.
  * Android Auto gains a toggleable, reorderable Lyrics section.
  * App auto-fetches and stores missing lyrics when available.
  * Shows placeholders for "no song playing" or "lyrics not found."
  * Displays synchronized lyrics as a 3-line current/previous/next view and keeps them in sync with playback.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MediaLibrarySessionCallback.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MediaLibrarySessionCallback.kt
@@ -61,9 +61,19 @@ import kotlinx.coroutines.plus
 import javax.inject.Inject
 import com.metrolist.music.constants.AndroidAutoSectionsOrderKey
 import com.metrolist.music.constants.AndroidAutoYouTubePlaylistsKey
+import com.metrolist.music.db.entities.LyricsEntity
+import com.metrolist.music.extensions.metadata
+import com.metrolist.music.lyrics.LyricsEntry
+import com.metrolist.music.lyrics.LyricsHelper
+import com.metrolist.music.lyrics.LyricsUtils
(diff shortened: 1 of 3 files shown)


johannesbrauer:
Didn't you noticed that we're in Maintenance mode and no new feature PR's will be merged?

isotjs:
> Didn't you noticed that we're in Maintenance mode and no new feature PR's will be merged?

I think this might count as a fix

johannesbrauer:
@isotjs I don't think so, they said that only bug fixes will be accepted, since that isn't a bug and doesn't affect the app usage it isn't a fix in my opinion.

FireLion137:
You're right to have doubts about this. I thought what I proposed wasn't a main feature since it was already present in the main app, but rather a visual addition to something that wasn't yet available on Android Auto. However, if you want, you can close it and I'll make a new PR when the addition is possible. No problem 👍

PR #3541, opened 2026-04-13
[wontfix] fix(readme): resize icon to 125px

kairosci:
## Problem
The Metrolist app icon in the README.md is 200px wide, which is larger than the text below it.

## Cause
The icon width was set to 200px while the heading text "Metrolist" and other text are sized differently.

## Solution
- Changed the icon width from 200px to 125px to match the text size


nyxiereal:
nah, this is actually what we want

PR #3558, opened 2026-04-14
fix: corrected the Build process of the play-next shuffle order

### LINKED ISSUE #2617: Queue starts from the beginning when it adding song to play nextshouldn't
### ⚠️ Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x]  This issue contains only ONE bug.

1)be in the middle of a shuffled queue
2)add song to play next
3)after the song chosen to play next finishes the queue will start from the top in the same order 
4)this causes you to relisten to the same queue again rather than continuing from where it is meant to be

queue has to be shuffled from the queue button not starting the playlist from shuffle

### Expected behavior

Instead the queue should continue from where it is even if song is added to play next 

### Actual behavior

this causes you to relisten to the same queue again rather than continuing from where it is meant to be

https://github.com/user-attachments/assets/5965095e-23fb-4fcb-b16f-6e1b96d609b7

### Logs

I have screen recording as evidence 

### Metrolist version

Still an issue with the latest 12.12.4 and previous versions 

### Android version

android 16 and android 12 

### Additional information

_No response_

johannesbrauer:
## Problem
When adding a song via "Play Next" with shuffle mode enabled, playback would restart the shuffle queue from the beginning instead of continuing from the current position.

## Cause
DefaultShuffleOrder is a full permutation array. ExoPlayer determines the current position within the shuffle order by finding currentMediaItemIndex in that array. The old code placed currentIndex at position 0 of finalOrder, which ExoPlayer interpreted as the start of the queue — effectively discarding all shuffle history and making it seem like playback restarted from scratch.

## Solution
Reconstruct finalOrder in the correct order: previously played items (prevList) first, then currentIndex, then the newly inserted items (nextBlock), then the remaining upcoming items (orderAfter). This preserves the shuffle history on both sides of the current position and correctly inserts the new items as immediate next-up without disrupting the rest of the queue. As a cleanup, the intermediate existingOrder variable (prevList + orderAfter) was removed since both lists are now used separately.

## Testing
- Build successful 
- Tested play-next with shuffle

## Related Issues
- Closes #2617 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shuffle behavior when using "Play Next": newly added tracks are now integrated into the shuffled queue so the current song, the inserted next block, and the remaining shuffled tracks play in the intended order, avoiding misplacement of pre-existing items.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -1671,9 +1671,10 @@ class MusicService :
                 val nextBlock = (insertIndex until (insertIndex + items.size)).toList()
                 val finalOrder = IntArray(size)
                 var pos = 0
+                prevList.forEach { if (it in 0 until size) finalOrder[pos++] = it }
                 finalOrder[pos++] = currentIndex
                 nextBlock.forEach { if (it in 0 until size) finalOrder[pos++] = it }
-                existingOrder.forEach { if (pos < size) finalOrder[pos++] = it }
+                orderAfter.filter { it !in newIndices }.forEach { if (pos < size) finalOrder[pos++] = it }
 
                 // Fill any missing indices (safety) to ensure a full permutation
                 if (pos < size) {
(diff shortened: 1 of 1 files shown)


alguezz:
When will this change be implemented?

nyxiereal:
@alguezz its in nightly and will be in v13.4.2

PR #3577, opened 2026-04-16
feat: migrate to Protobuf Gradle plugin for OS-independent generation

kairosci:
## Problem
Proto generation was manual and OS-dependent, requiring manual installation of protoc on the developer machine.

## Cause
The build system relied on local shell scripts and OS-specific binary downloads within the build script.

## Solution
This PR integrates the official Protobuf Gradle plugin for seamless and OS-independent proto generation. It configures the plugin to use lite java and kotlin variants for Android. The proto file is now tracked within the app module for better self-containment.

## Testing
The changes were verified by running the generateFossDebugProto task and successfully compiling the project using the compileFossDebugKotlin task. Generated files were inspected in the build directory to ensure correctness.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated protobuf code generation into the Gradle build system, removing it from separate CI/CD workflow steps. Updated build configuration to streamline the build process while maintaining all existing functionality and dependencies.


nyxiereal:
I did the initial protobuf related things in the app. You don't need to mention mo

.github/actions/setup-protobuf/action.yml
nyxiereal, line 1: this action just becomes useless. you can simply add it to the build.gradle.kts file, so its ran automatically (without actions steps). 

you can then edit every workflow, and finally remove `.github/actions/setup-protobuf/action.yml`

gradle/libs.versions.toml
```diff
@@ -36,6 +36,7 @@ kuromojiIpadic = "0.9.0"
 extractor = "6305155"
 tinypinyin = "2.0.3"
 protobuf = "4.34.1"
+protobufPlugin = "0.9.4"
```
nyxiereal: use 0.9.6 unless you have a good reason not to

https://plugins.gradle.org/plugin/com.google.protobuf

gradle.properties
```diff
@@ -17,6 +17,7 @@ android.useAndroidX=true
 android.enableJetifier=false
 org.gradle.unsafe.configuration-cache=true
 android.nonTransitiveRClass=false
+android.newDsl=false
```
nyxiereal: why?

nyxiereal (approved):

nyxiereal:
i didnt mean to approve, missclicked

gradle.properties
kairosci: Without Gradle fails, I did not understand why, I will try to understand it.

kairosci:
I made `protobuf` generation a build dependency. 

I also slightly changed my approach, because using protobuf directly required an AGP version bump, since the current version was causing compilation errors.

So it installs it as a third-party Gradle dependency and runs it to generate the file.

I’ve achieved my goal (making the generation cross-platform); it seems to build and run correctly.

nyxiereal (approved):

PR #3636, opened 2026-04-24
fix(notification): prevent cancel action duplication in download notification

isotjs:
## Problem
The cancel button in the download foreground notification multiplies itself on each update (e.g., "cancel", "cancel cancel", "cancel cancel cancel").

## Cause
`getForegroundNotification()` in `ExoDownloadService` calls `addAction()` on a reused `Notification.Builder`. `DownloadService` calls `getForegroundNotification()` every second via its `ForegroundNotificationUpdater`. Each call adds another cancel action on top of the existing ones, causing the label to repeat.

## Solution
- Build the base progress notification first, then use `setActions()` (which replaces all actions) instead of `addAction()` (which accumulates)
- Behavior unchanged (cancel removes all pending downloads)

## Testing
- Tried on emulator and S25 fixed fine.

## Related Issues
- Related to Thread id 1497149279859376149 on discord.

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal notification handling for improved code structure and maintainability.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/ExoDownloadService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/ExoDownloadService.kt
@@ -50,18 +50,22 @@ class ExoDownloadService : DownloadService(
     override fun getForegroundNotification(
         downloads: MutableList<Download>,
         notMetRequirements: Int
-    ): Notification =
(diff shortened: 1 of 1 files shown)


nyxiereal:
Read "thread id" as "threat id"

nyxiereal (approved):

PR #3646, opened 2026-04-25
Fix playback issues in 13.4.2

gergesh:
Do not merge yet (well, do whatever you want but) this is just a draft PR so an official debug build becomes available

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of function-name extraction to skip non-matching patterns and avoid extraction errors; added debug logging when a pattern is ignored.
  * Prevented potential hangs during token generation by enforcing a generation timeout, performing graceful cleanup on timeout, and falling back so the app can continue without blocking.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/cipher/FunctionNameExtractor.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/cipher/FunctionNameExtractor.kt
@@ -273,6 +273,13 @@ object FunctionNameExtractor {
                         return NFunctionInfo(name, arrayIdx, isHardcoded = false)
                     }
                     else -> {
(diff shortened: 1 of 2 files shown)


gergesh:
Hey @nyxiereal 
1. I am unsure if the second commit here (the timeout-related one) is even important. I think the first fixes the issue.
2. If the first fixes the issue, I guess the question is why does the first parser fail and we fallback to the VR one. My commit makes the crash recoverable, but it might be possible to avoid it entirely?

nyxiereal:
1. maybe? i think having a timeout is somewhat important, BUT can cause issues for people with shitty network conditions
2. i dont know, i didnt really work on the potoken/player stuff that much

i will fix my issue in a sec

PR #3651, opened 2026-04-26
fix(player): add age-restricted playback error string resource

Anuj3937:
## Problem
PlaybackError.kt now references R.string.error_age_restricted, but metrolist_strings.xml did not define that string resource.

## Cause
The playback error message was moved to resources without adding the corresponding error_age_restricted entry.

## Solution

- Added stringResource(R.string.error_age_restricted) usage in PlaybackError.kt
- Added <string name="error_age_restricted">This app does not support playing age-restricted songs. We are working on fixing this issue.</string> to metrolist_strings.xml

## Testing
- ./gradlew :app:assembleFossDebug passed successfully
- Confirmed metrolist_strings.xml is syntactically valid and contains the error_age_restricted string resource with correct text.
- Code Reference Validation: Verified PlaybackError.kt line 57 contains stringResource(R.string.error_age_restricted) and the resource name follows Android naming conventions.

## Related Issues

- Closes #
- Related to #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messaging for age-restricted content with proper localization support across different languages.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/player/PlaybackError.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/player/PlaybackError.kt
@@ -54,7 +54,7 @@ fun PlaybackError(
             error.errorCode == PlaybackException.ERROR_CODE_IO_BAD_HTTP_STATUS
     
     val errorMessage = if (isAgeRestricted) {
-        "This app does not support playing age-restricted songs. We are working on fixing this issue."
+        stringResource(R.string.error_age_restricted)
     } else {
         rawErrorMessage
     }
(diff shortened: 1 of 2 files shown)


johannesbrauer:
Idk if this is helpful, could make it harder for us to analyze the error if the app language is different than English.

Anuj3937:
> Idk if this is helpful, could make it harder for us to analyze the error if the app language is different than English.

well it was hardcoded , it is better like this tbh, follows the same pattern as other existing errors

johannesbrauer:
> > Idk if this is helpful, could make it harder for us to analyze the error if the app language is different than English.
> 
> well it was hardcoded , it is better like this tbh, follows the same pattern as other existing errors

You're completely right, but people are just too lazy to change there app language when they report a bug, so it makes it harder for us to understand the problem. But you're right it's important to keep things consistent around the whole app.

It's just my opinion.

nyxiereal:
nope, the support team operates only in english.

PR #3657, opened 2026-04-27
Enhance experimental lyrics : fix devnagari rendering glitch

cloud-zip:
## Problem
Devanagari and other Indic script lyrics render with broken and fragmented glyphs in the expressive word-level lyrics view. Vowel matras and combining marks detach from their base consonants during animation, causing visual corruption.

## Cause
The animation pipeline iterated over codepoint indices (for i in mainText.indices), treating each Char as an independent unit. Indic glyphs like में are 3 codepoints but 1 visible cluster, so each combining mark got its own bounding box, scale transform, and draw call, tearing them apart visually.

## Solution
- Added String.toGraphemeClusters() using java.text.BreakIterator.getCharacterInstance() to correctly segment text into visual grapheme units
- Rebuilt charToWordData to be indexed by cluster count instead of codepoint count
- Added clusterCharOffsets to map cluster indices back to String/Char offsets required by TextLayoutResult methods
- Changed letterLayouts to measure one full cluster string per entry instead of one Char
- Replaced pre-calc and main drawing loops to iterate over cluster indices (0 until clusterCount)
- Updated RTL word-bounds scan to use cluster indices and clusterCharOffsets for correct glyph offset lookups
- Renamed clusterCodePointOffsets to clusterCharOffsets with clarified comments stating String offsets are UTF-16/Char indices

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lyrics rendering and animation so complex characters, emoji, and multi-script text (e.g., Devanagari, Bengali, Arabic) render as single visual units. Alignment, clipping, highlighting, wobble and glow effects now apply correctly across LTR and RTL text, eliminating visual fragmentation and misaligned bounding boxes during playback.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/component/LyricsLine.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/component/LyricsLine.kt
@@ -91,6 +91,27 @@ private fun String.containsRtl(): Boolean {
     return false
 }
 
+/**
+ * Splits a string into Unicode grapheme clusters using BreakIterator.
+ * This correctly handles Devanagari, Bengali, Arabic, Hangul, emoji, etc.
+ * where a single visible glyph is composed of multiple code points (e.g. base
+ * consonant + matra + anusvara = one cluster, not three separate chars).
+ */
+private fun String.toGraphemeClusters(): List<String> {
+    if (isEmpty()) return emptyList()
+    val result = mutableListOf<String>()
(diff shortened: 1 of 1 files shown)


cloud-zip:
Chinese and Japanese working, tested.

nyxiereal (approved):
LGTM

cloud-zip:
> LGTM

Uhm didn't understand?

nyxiereal:
LGTM means looks good to me

PR #3695, opened 2026-05-03
fix(artists): properly handle artist conjunctions in YouTube Music API responses

kairosci:
## Problem

Songs appear under wrong artist pages because the YouTube Music API returns multiple artists joined by conjunctions (e.g., "Artist1 & Artist2", "Artist1 y Artist2", "Artist1 e Artist2") as a single run. The app was treating these as a single artist, causing songs to appear only under one artist instead of all relevant ones.

## Cause

The artist parsing logic in `ArtistPage.kt`, `ArtistItemsPage.kt`, `SearchSummaryPage.kt`, and `SearchSuggestionPage.kt` did not split artist runs by conjunction patterns. When artists like "ditonellapiaga e tonypitony" were returned, the app created one artist "ditonellapiaga e tonypitony" instead of two separate artists.

## Solution

- **`Runs.kt`** — Added `splitArtistsByConjunction()` that splits runs on localized conjunction words (space-padded) and ` & `. Comma (`,`) is NOT used as separator to avoid breaking names like "Tyler, The Creator". NavigationEndpoint is shared to all split parts so no artist loses its ID. Updated `clean()` to retain groups with localized conjunction patterns.
- **`Runs.kt`** — `ArtistConjunctions` singleton holds configurable word list, defaulting to `listOf("and")`.
- **`App.kt`** — Loads conjunctions from `metrolist_strings.xml` via locale-aware resources. The default file keeps only `"and"`; localized conjunctions (`e`, `y`, `et`, `und`, etc.) are added in their respective language files.
- **Page parsers** — Applied `splitArtistsByConjunction()` in `ArtistPage.kt`, `ArtistItemsPage.kt`, `SearchSummaryPage.kt`, `SearchSuggestionPage.kt`. Two-row fallback now preserves **all** split artists with null IDs instead of collapsing to `firstOrNull()`.
- **`YouTube.kt`** — Added `resolveArtistIds()` that searches for missing artist IDs and assigns them only on exact name match.

## Key fixes

1. `, ` removed from separator pattern — prevents splitting "Tyler, The Creator"
2. `navigationEndpoint` shared to all split parts — no null IDs after split
3. `clean()` checks ALL runs in first group (not just first) — localized conjunctions not dropped
4. Two-row fallback uses `artistRuns.map` instead of `firstOrNull()` — no artist lost
5. `resolveArtistIds` uses exact name match — prevents wrong artist IDs
6. Default `metrolist_strings.xml` keeps only `"and"` — other conjunctions go in locale files

## Testing

- Builds successfully with `./gradlew :app:assembleFossDebug`
- innertube module compiles with `./gradlew :innertube:compileDebugKotlin`
- Need manual testing on device with various conjunction patterns

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/App.kt
+++ b/app/src/main/kotlin/com/metrolist/music/App.kt
@@ -97,16 +97,6 @@ class App :
 
         ArtistConjunctions.conjunctions = listOf(
             R.string.and,
-            R.string.e,
-            R.string.ed,
-            R.string.y,
-            R.string.et,
-            R.string.und,
-            R.string.en,
-            R.string.och,
-            R.string.og,
-            R.string.ja,
-            R.string.ve,
         ).mapNotNull { id ->
             runCatching { getString(id) }.getOrNull()
         }
--- a/app/src/main/kotlin/com/metrolist/music/App.kt
+++ b/app/src/main/kotlin/com/metrolist/music/App.kt
@@ -22,6 +22,7 @@ import coil3.request.CachePolicy
 import coil3.request.allowHardware
 import coil3.request.crossfade
 import com.metrolist.innertube.YouTube
+import com.metrolist.innertube.models.ArtistConjunctions
 import com.metrolist.innertube.models.YouTubeLocale
 import com.metrolist.kugou.KuGou
 import com.metrolist.lastfm.LastFM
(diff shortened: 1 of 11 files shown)


kairosci:
[WIP]: I HAVEN'T TEST YET.

kairosci:
It seems to work

PR #3780, opened 2026-05-22
Fix the bug on the Xiaomi hyper OS where the album gets stuck loading when…

Arjuanto:
Bug

  On the Xiaomi Hyper OS (Indonesian locale, without a VPN), when opening any album page in Metrolist:
  - The album page opens successfully (album title and thumbnail appear)
  - The track list never appears — the spinner keeps spinning indefinitely
  - No error message, no crash
  - Reproduces 100% of the time on all albums, from any source (Library / Search / Home / Artist)

  Root Cause

  The bug is located in innertube/src/main/kotlin/com/metrolist/innertube/utils/Utils.kt:94 within the String.parseTime() function:

  fun String.parseTime(): Int? {
      try {
          val parts = split(“:”).map { it.toInt() }   // ← split only by “:”
          if (parts.size == 2) return parts[0] * 60 + parts[1]
          ...

  What happens step-by-step:

  1. The app fetches the album via YouTube.album(browseId) → JSON response from YouTube Music
  2. The response contains 24 songs with the duration field as a string
  3. YouTube Music sends the duration based on the device’s locale. Locale en-US → “4:25”. Locale ID/EU → “4.25” or “4,25”.
  4. parseTime(“4.25”) → split(“:”) → results in [“4.25”] (size 1, not 2/3) → returns null
  5. In AlbumPage.getSong() line 108:
  duration = renderer...?.text?.parseTime() ?: return null
  5. parseTime returns null → the entire getSong() returns null → the song is discarded
  6. mapNotNull { getSong(it, album) } → all 24 items are discarded → empty list
  7. AlbumPage returns with songs = [] → the DB is updated with 0 songs → the UI keeps loading because the albumWithSongs flow doesn’t emit a list of songs

  What hides this bug:
  - No exception — parseTime silently returns null, swallowed by ?: return null
  - mapNotNull discards null without logging
  - runCatching {} in YouTube.album() wraps everything as “success” (since there’s no throw, just an empty list)
  - reportException is never called → clean logcat
  - On emulator/device with en-US locale → cannot reproduce due to duration using :

  Fix

  One line in Utils.kt:96:

  // Before
  val parts = split(“:”).map { it.toInt() }

  // After
  val parts = split(Regex(“[:.,]”)).map { it.toInt() }

  Now parseTime accepts :, ., and , as separators. Formats like “4.25”, “4:25”, and “4,25” all result in 265 seconds (4 minutes × 60 + 25). The time format “1:23:45” also works as expected.

  Since parseTime is used in all duration parsers (album, playlist, search, artist page), this single fix also resolves song durations that previously displayed as 0:00 in other areas—though the issue isn’t as critical   
  as with albums (it’s merely cosmetic).

  How to reproduce (from start to confirmation)

  Prerequisites: Device/emulator with a non-en-US locale (e.g., Indonesia / Germany) without a VPN. Metrolist debug build installed.

  1. Open the Metrolist app
  2. Open any YouTube Music album (Library / Search / Home / Artist page)
  3. Symptom: Spinner spins, song never appears, no error

  Quick way to verify root cause without a rebuild (on emulator or via adb shell):
  adb logcat -c
  adb logcat AlbumVM:V *:S
  Tap the album → if “albumSongs: rawItems=N, parsedSongs=0” appears with N > 0 → confirms the parser is discarding all songs. (Note: that log was present in the instrumented build — I removed it after the fix)        

  After applying the fix:
  1. Rebuild: gradlew assembleGmsDebug
  2. Install: adb install -r app\build\outputs\apk\gms\debug\app-gms-debug.apk
  3. Open the same album
  4. Expectation: 24 songs appear with correct durations (e.g., “4:25”), the spinner disappears, and songs can be played

  Lessons

  Bugs like this easily slip through testing because:
  - CI typically runs on US-locale servers → never triggers
  - Silent logging (?: return null, mapNotNull, runCatching) is intentionally designed for graceful degradation but ends up hiding real bugs
  - YouTube Music’s per-locale behavior isn’t officially documented → developers don’t anticipate that separators might differ

  Defensive parsing for external data (especially from frequently changing APIs like YouTube InnerTube) should always accept multiple formats and log a warning when a fallback is used, so bugs can be detected
   more quickly.

Translated with [DeepL.com](https://www.deepl.com/?utm_campaign=product&utm_source=web_translator&utm_medium=web&utm_content=copy_free_translation) (free version)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed time duration parsing to support additional time format separators from YouTube Music, ensuring proper display across different locales.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MetrolistGroup/Metrolist/pull/3780?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nyxiereal (approved):

nyxiereal:
@adrielGGmotion can you verify, you have hyperos if i remember correctly

adrielGGmotion:
> @adrielGGmotion can you verify, you have hyperos if i remember correctly

Nyx i never touched a HyperOS device on my life 😂

Arjuanto:
@nyxiereal @adrielGGmotion This is because I personally use a Xiaomi cellphone and most Xiaomi cellphones now use Xiaomi hyperOS, not only Xiaomi that uses HyperOS, there are also several other cellphones such as Redmi and Poco

Arjuanto:
And in my country, many people also use that device.

moonpic:
Bug verification:

https://github.com/user-attachments/assets/c7526fbb-3fea-4848-92f8-68bcc68f8c42

Fix verification:

https://github.com/user-attachments/assets/1bf73a9c-bb16-4e24-b398-aa7eedf9f7bf

PR #3825, opened 2026-05-27
fix(ui): Fix discord integration's Activity name field validation

inlinefun:
## Problem

No way to set the discord integration's "Activity name" field to be an empty value

## Cause

The Dialog informs the user that the text field can be left empty to use the default value. The text field would try to evaluate if the text is empty or not, which is not possible due to input validation

## Solution

- Change the input validation to always be true, even when empty

## Testing

- Successfully compiled the application
- The feature now works as intended, screenshots attached

Before the change
image
After the change
image

## Related Issues

- None, could not find any

## Summary by CodeRabbit

* **Bug Fixes**
  * Discord activity name input now trims leading/trailing whitespace and reliably accepts entries; the dialog’s Done action is consistently enabled so names are saved as entered, preventing accidental blank or whitespace-only activity names.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MetrolistGroup/Metrolist/pull/3825?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kairosci (approved):

nyxiereal:
not needed after the rewrite

PR #3837, opened 2026-05-28
feat(ui): add predictive back support to BottomSheet

HansHolz09:
## Problem
The player queue overlays currently appear only slightly transparent over scrolling content, which I personally don't find very visually appealing. The player and queue close only after the back action completes instead of following predictive back progress.

## Cause
The queue header and bottom controls use translucent backgrounds without a blurred backdrop layer behind them. The bottom sheet still uses a regular back handler, so it cannot react to predictive back gesture progress.

## Solution
- Add a reusable overlay backdrop blur modifier for Android 12+
- Apply backdrop blur behind the queue header and bottom controls
- Keep pure black mode and older Android versions on the existing non-blurred fallback
- Add predictive back handling to the bottom sheet so the player and queue sheet follow the back gesture

## Testing
- Compiled and testet

https://github.com/user-attachments/assets/da41b3a9-0b67-46c9-bcec-52a1d3d05b41

## Summary by CodeRabbit

* **Improvements**
  * Enhanced back gesture handling with smooth predictive animations that respond dynamically to user input. The app now provides improved visual feedback during back gestures with better state management for cancelled navigation actions.

* **Style**
  * Minor code formatting updates for improved readability in the queue view.


AMiguel7:
personally it'd be better if they'd be removed completely and in place of that we provide an image blured to 100-200 px surface blur image.png of custom for users to choose to be background of whole app.. besides the ui here would be better like this !

nyxiereal:
> personally it'd be better if they'd be removed completely and in place of that we provide an image blured to 100-200 px surface blur image.png of custom for users to choose to be background of whole app.. besides the ui here would be better like this ! 

this looks like vibecoded garbage, not following material 3 design standards

nyxiereal:
@HansHolz09 hey thanks for the addition, but could you just remove the blur from the backgrounds. it doesnt fit with material design - which say to limit blur

app/src/main/kotlin/com/metrolist/music/ui/utils/OverlayBackdropBlur.kt
nyxiereal, line 1: pls no blur thx

app/src/main/kotlin/com/metrolist/music/ui/player/Queue.kt
```diff
@@ -133,14 +133,14 @@ import sh.calvin.reorderable.ReorderableItem
...
-import androidx.datastore.preferences.core.edit
 import android.widget.Toast
+import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.runtime.derivedStateOf
 import com.metrolist.music.constants.SleepTimerFadeOutKey
 import com.metrolist.music.constants.SleepTimerStopAfterCurrentSongKey
-import androidx.compose.runtime.derivedStateOf
 import androidx.compose.material3.Button
+import androidx.compose.ui.graphics.rememberGraphicsLayer
+import com.metrolist.music.ui.utils.overlayBackdropBlur
```
nyxiereal: yeah

nyxiereal (changes requested):

HansHolz09:
> @HansHolz09 hey thanks for the addition, but could you just remove the blur from the backgrounds. it doesnt fit with material design - which say to limit blur

I added the blur because Google seems to be leaning more into it with Material 3 Expressive, especially in the newer Pixel UI (https://9to5google.com/2026/05/19/android-17-beta-1-blur/). Could we at least remove the alpha / slight transparency instead, since that feels even less aligned with M3E? Otherwise, I can reduce this PR to only the predictive back changes.

app/src/main/kotlin/com/metrolist/music/ui/utils/OverlayBackdropBlur.kt
AMiguel7: i mean if you don't like it have it on only by certain toggle.. in settings below dynamic icon option

AMiguel7:
> > personally it'd be better if they'd be removed completely and in place of that we provide an image blured to 100-200 px surface blur image.png of custom for users to choose to be background of whole app.. besides the ui here would be better like this ! 
> 
> this looks like vibecoded garbage, not following material 3 design standards

i meant for the layout not background go on layout

nyxiereal:
> I can reduce this PR to only the predictive back changes.

fair, sure - do that

PR #3807, opened 2026-05-25
fix(audio): custom volume normalization to resolve muffled audio

kairosci:
## Problem
Audio frequently sounds muffled.

## Cause
The use of Android's native `LoudnessEnhancer` for volume normalization triggers the hardware DSP on many devices, applying aggressive dynamic range compression that squashes high frequencies and muffles the audio even when simply normalizing.

## Solution
- Replaced `LoudnessEnhancer` with a custom `VolumeNormalizationAudioProcessor`.
- The new processor applies pure linear gain directly to 16-bit PCM audio samples inside the ExoPlayer chain.
- Bypassed the Android OS audio effects pipeline completely to preserve pristine audio quality.
- Removed all `LoudnessEnhancer` session management logic.

## Testing
- Successfully compiled `assembleFossDebug`.
- Verified audio normalization path bypasses LoudnessEnhancer and correctly handles gain mathematically.

## Related Issues
- Fixes #3833 

## Summary by CodeRabbit

* **New Features**
  * Per-track volume normalization applied immediately from cached settings and updated asynchronously for smoother playback.
  * Normalization now runs earlier in the audio pipeline, supports common PCM/float encodings, and offers enable/disable plus configurable target gain.

* **Bug Fixes**
  * Safer error handling and simplified audio-session tracking: failures now disable normalization without disrupting playback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MetrolistGroup/Metrolist/pull/3807?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3859, opened 2026-06-01
[wontfix] fix(db): add manual migration from 38 to 37 to prevent data loss on downgrade

kairosci:
## Problem
While testing PRs or branches containing a database schema bump to version 38 (such as PR 3807), I noticed a fatal `IllegalStateException` when returning to `main` or downgrading to version 37. Room crashes because it requires a downgrade migration path that doesn't exist. Using destructive migration fallback on downgrade is not an option as it results in total user data loss.
Furthermore, I discovered that the dependency injection in `AppModule.kt` was instantiating the Room database using a bare `Room.databaseBuilder`, completely bypassing `InternalDatabase.newInstance(context)`. This caused ALL manual database configurations, migrations, and callbacks to be ignored by the app.

## Cause
- When I upgrade the Room database to version 38 and then return to an older build running database version 37, it triggers a downgrade path. Since no downgrade migration was provided in `main`, Room throws an exception: `A migration from 38 to 37 was required but not found.`
- `AppModule.kt` was ignoring `MusicDatabase.newInstance()` which meant that even if a migration was added there, Room would never see it.

## Solution
- I fixed `AppModule.kt` to actually use `InternalDatabase.newInstance(context)` and properly extract the delegate, so all manual migrations and configurations are correctly applied.
- I added a manual downgrade migration `MIGRATION_38_37` inside `MusicDatabase.kt`.
- My migration safely drops the `ArtistPageCache` table (added in 38) and gracefully reverts the `artist` table structure to its schema at version 37 by reconstructing the table.
- I removed the destructive migration fallback on downgrade to guarantee no user data is wiped when downgrading.

## Testing
- I verified that `MusicDatabase.kt` compiles successfully and the SQL matches the Room schema for version 37 exactly.
- I ran `./gradlew :app:assembleFossDebug` to ensure `ksp` verification passes successfully on my end.
- The migrations are now correctly registered and executed.

## Related Issues
- Related to #3807

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artist information handling and data management to enhance app stability.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/db/MusicDatabase.kt
+++ b/app/src/main/kotlin/com/metrolist/music/db/MusicDatabase.kt
@@ -177,6 +177,7 @@ abstract class InternalDatabase : RoomDatabase() {
                             MIGRATION_21_24,
                             MIGRATION_22_24,
                             MIGRATION_24_25,
+                            MIGRATION_38_37,
                         ).fallbackToDestructiveMigration(dropAllTables = true)
                         .setJournalMode(RoomDatabase.JournalMode.WRITE_AHEAD_LOGGING)
                         .setTransactionExecutor(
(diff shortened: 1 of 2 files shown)


nyxiereal:
no

nyxiereal:
downgrades are explicitly NOT supported by the app, and i fucking hate it when people downgrade the app and then bitch about it bricking itself

kairosci:
understood, then sorry for pr.

PR #3890, opened 2026-06-03
fix(ui): correct FAB position in search screen

il-Luca:
## Problem

On the search screen, the music-recognition FAB sits higher than on every other screen (Home, History, Library). The offset is more noticeable on some devices (e.g. Pixel 9a) than others.

Media (1) 
Media

## Cause

The search screen is the only screen that hosts the FAB inside a `Scaffold`. Its content `Box` applied the full `paddingValues` from the Scaffold (which includes the bottom system-bar inset) and wrapped the content in an extra inner `Box` with `padding(bottom = bottomPadding)`. The `HideOnScrollFAB` already accounts for the bottom inset internally via `LocalPlayerAwareWindowInsets`, so the Scaffold's bottom padding was applied twice, pushing the FAB up. Other screens don't use a `Scaffold`, so they never hit this double-padding.

## Solution

- Apply only the top padding from the Scaffold's `paddingValues` (needed for the search bar app bar), letting `HideOnScrollFAB` handle the bottom inset itself, consistent with the other screens.
- Remove the redundant inner `Box` and its `bottomPadding`, so the search content and the FAB are direct siblings in a single `Box`.

## Testing

- Built with `./gradlew :app:assembleFossDebug`.
- Verified on physical devices that the FAB now sits at the same height as on Home/History/Library, with 3-button navigation.

## Related Issues

MInor cosmetic bug, no issue reported.

## Summary by CodeRabbit

* **Refactor**
  * Simplified the search screen's layout and padding structure for cleaner rendering and maintainability; no changes to visible behavior.
  * Preserved existing dismiss behaviors: online search remains suppressed during suggestion searches, and local search still navigates back as before.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/search/SearchScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/search/SearchScreen.kt
@@ -279,40 +279,31 @@ fun SearchScreen(
         },
         containerColor = if (pureBlack) Color.Black else MaterialTheme.colorScheme.background,
     ) { paddingValues ->
-        val bottomPadding = LocalPlayerAwareWindowInsets.current.asPaddingValues().calculateBottomPadding()
-
         Box(
             modifier =
                 Modifier
-                    .padding(paddingValues)
+                    .padding(top = paddingValues.calculateTopPadding())
                     .fillMaxSize(),
         ) {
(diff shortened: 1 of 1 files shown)


il-Luca:
When switching to search screen the FAB button is hidden by the keyboard. I did not change that behavior.

PR #3889, opened 2026-06-03
fix(recognition): use non-deprecated flush() to solve "failed to resample audio" issue

il-Luca:
## Problem

Music recognition fails with the error "Failed to resample audio: AudioProcessor must implement at least one #flush() overload."

## Cause

`AudioResampler` calls the deprecated no-argument `AudioProcessor.flush()` on a `SonicAudioProcessor`. In Media3 1.10.x the `AudioProcessor` interface changed: the no-arg `flush()` is now a deprecated default method that throws `IllegalStateException`, and implementations are expected to use the new `flush(StreamMetadata)` overload instead.

The recognition feature was originally written against Media3 1.7.1 (where the old `flush()` worked). When the project bumped Media3 to 1.10.x, `AudioResampler` was not updated to the new API, so the call now hits the throwing default.

## Solution

- Replace the deprecated `sonic.flush()` call with `sonic.flush(AudioProcessor.StreamMetadata.DEFAULT)`, the non-deprecated overload that correctly invokes `SonicAudioProcessor`'s implementation.

## Testing

- Built with `./gradlew :app:assembleFossDebug` (clean build).
- Verified on a physical device: music recognition now records, resamples, and returns a match successfully where it previously failed with the resample error.

## Related Issues

- Closes https://github.com/MetrolistGroup/Metrolist/issues/3834 , https://github.com/MetrolistGroup/Metrolist/issues/3766

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio stream handling during music recognition to ensure proper metadata processing at stream completion.

PR #3891, opened 2026-06-03
feat(recognition): simplify recognition result screen

il-Luca:
## Problem

The recognition result screen presented three buttons — "Play on App", "Re-listen", and "Close" — but they didn't work as intended or were redundant:

- **"Play on App"** couldn't reliably play the recognized song. It depended on a YouTube video ID from the Shazam response, but that field is almost never present, so the button fell back to a plain search instead of playing. Its label was therefore misleading.
- **"Re-listen"** was redundant: the result screen already returns to the listening state on back, so a dedicated button added nothing.
- **"Close"** duplicated the existing top-left back button / system back, offering no extra value.

There was also a navigation issue: going to a child screen (search results or recognition history) and pressing back lost the result, dropping the user on the "tap to recognise" state instead of the result they were viewing.

This is a usability improvement, not a new feature — it removes broken/redundant controls and fixes incorrect back behavior.

## Cause

- The Shazam recognition response only optionally includes a YouTube link in `hub.options`, and in practice it's almost always absent. "Play on App" relied on that ID, so it could rarely (if ever) actually start playback and silently degraded to a search.
- The button set included two controls ("Re-listen", "Close") that overlapped with existing back navigation.
- The recognition screen is a single destination driven by a shared status flow; returning from a child screen wasn't distinguished from a fresh entry, so the screen reset its state at the wrong time.

## Solution

- Replace the three buttons with a single **Search** button with a clear, accurate label.
- The Search button navigates to YouTube Music search using the recognized `title + artist`, which is the reliable way to find the track (no dependency on a Shazam-provided YouTube ID).
- Preserve the result state when navigating to child screens (search, recognition history) so back returns to the result.
- Reset to the "tap to recognise" state only on a fresh entry from Home/Search, not when returning from a child screen.

## Testing

- Verified on a physical device:
  - Recognize a song → result shows with a single Search button.
  - Tap Search → YouTube Music search opens with the correct query; back returns to the result (not the listening screen).
  - Open recognition history from the result and press back → result is still shown.
  - Opening recognition fresh from Home/Search shows the "tap to recognise" state.

## Summary by CodeRabbit

* **Improvements**
  * Back button now resets recognition status when results are displayed
  * Success screen features a dedicated "search" action to explore results further
  * History saving enhanced to prevent duplicate entries
  * Refined recognition and permission flow handling

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/recognition/RecognitionForegroundService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/recognition/RecognitionForegroundService.kt
@@ -264,7 +264,7 @@ class RecognitionForegroundService : Service() {
             contentIntent = pendingIntent,
             largeIcon = null,
             actionIntent = pendingIntent,
-            actionTitle = getString(R.string.listen_on_metrolist),
+            actionTitle = getString(R.string.recognition_open),
         )
 
         serviceScope.launch {
@@ -286,7 +286,7 @@ class RecognitionForegroundService : Service() {
                     contentIntent = pendingIntent,
                     largeIcon = coverBitmap,
                     actionIntent = pendingIntent,
-                    actionTitle = getString(R.string.listen_on_metrolist),
+                    actionTitle = getString(R.string.recognition_open),
                 )
             }
             finishWithPersistentResult()
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/recognition/RecognitionScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/recognition/RecognitionScreen.kt
@@ -44,7 +44,6 @@ import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.Card
 import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.FilledTonalButton
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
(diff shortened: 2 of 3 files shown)


il-Luca:
This is admittedly a quick and somewhat "lazy" fix. However, there may be better approaches to fetch the YouTube video ID using the Shazam response data – since the response itself doesn't seem to be returning the YouTube ID. This would allow for tighter integration with the app, enabling users to play the recognised song – or add it to their library – immediately after identification.

For reference, see how @aleksey-saenko handles this in MusicRecognizer: https://github.com/aleksey-saenko/MusicRecognizer/blob/master/core/metadata/src/main/java/com/mrsep/musicrecognizer/core/metadata/tracklink/youtube/YoutubeTrackLinksFetcher.kt

il-Luca:
>> [@adrielGGmotion](https://github.com/adrielGGmotion) FYI there is a PR https://github.com/MetrolistGroup/Metrolist/pull/3889 waiting to be merged since last week which fixes the same issue by changing 1 line of code 🙃. I am now closing it because your solution might be "cleaner". However, since you are working on the "music recognition" feature, maybe have a look at the other 2 PR I've opened and see if they make sense: https://github.com/MetrolistGroup/Metrolist/pull/3890 and https://github.com/MetrolistGroup/Metrolist/pull/3891
> >
> > ⚠️ I've "vibe-coded" them and I have little experience contributing to Android projects, so feel free to challenge/reject them
> 
> Thanks, I merged https://github.com/MetrolistGroup/Metrolist/pull/3890, but I didn't merge the other one because the function of the buttons is to keep you on the same screen and not exit it!

@mostafaalagamy no problem, I will close this PR to avoid adding to the list of ~20 PRs which are still open. Whenever you find the time, you should re-check the logic of the "recognize" feature because it doesn't make much sense to me. IMHO both the play button (not functional) and the close button (redundant) are problematic.

PR #3893, opened 2026-06-03
fix(media-notification): optimistic update for media notification like/save state

stopper2408:
## Problem
There was a second debounce before the UI updated for the media notification
## Cause
There was a 1 second debounce
## Solution
I have implemented an optimistic update to make it instant

## Testing
Built the app, then tested on my Pixel phone

## Related Issues
N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Notification and widget UI now update instantly when you like or save songs and episodes, providing real-time visual feedback for your actions without delay.


stopper2408:
@coderabbitai review

PR #3935, opened 2026-06-06
[wontfix] fix(sync): append offline liked songs to youtube instead of overwriting

### LINKED ISSUE #3685: Local liked songs overwritten by YouTube account liked songs
### Checklist - Required

1. Like songs without being logged into a YouTube account.
2. Log into a YouTube account.
3. Check "Liked".

### Expected behavior

The local liked songs shouldn't be overwritten. It would make sense if the YouTube account liked songs got appended to the list.

### Actual behavior

The local liked songs are fully replaced with the YouTube account ones.

_No response_

### Logs

No logs

### Metrolist version

13.4.2

### Android version

11

### Additional information

_No response_

kairosci:
## Problem
Local liked songs are getting overwritten by the YouTube account liked songs when a user logs in, effectively wiping out their offline likes.

## Cause
The sync logic in `SyncUtils` was removing local likes for any songs not present in the remote YouTube list, without accounting for songs that were liked while the user was offline or prior to logging in.

## Solution
- Checked the `LastFullSyncKey` timestamp to determine if a local like was added after the last sync or while offline.
- Pushed offline likes to YouTube during sync, appending them to the account instead of overwriting them.
- Ignored local device files (`isLocal`) during the sync so they are never unliked by the YouTube remote state.

## Testing
Analyzed and tested the sync algorithm to ensure offline likes are preserved and pushed, while still respecting unlikes made remotely.

## Related Issues
- Closes #3685

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization of liked songs between local and remote libraries: local-only likes are no longer removed during cleanup. Instead, qualifying local likes are pushed to the remote list.
  * Tightened merge/update behavior to apply changes more selectively, helping preserve existing like timestamps when appropriate and keeping video status aligned with the remote source.
  * Enhanced completion reporting to include both remote and local liked song counts for clearer sync results.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
@@ -734,10 +734,17 @@ class SyncUtils @Inject constructor(
                     val remoteIds = remoteSongs.map { it.id }.toSet()
                     val localSongs = database.likedSongEntitiesByNameAsc()
 
+                    val lastSync = context.dataStore.get(LastFullSyncKey, 0L)
+
                     // Remove likes from songs not in remote
-                    localSongs.filterNot { it.id in remoteIds }.forEach { song ->
+                    localSongs.filterNot { it.id in remoteIds || it.song.isLocal }.forEach { song ->
                         try {
-                            database.update(song.localToggleLike())
+                            val likedDate = song.song.likedDate
+                            if (likedDate != null && likedDate.toEpochSecond(ZoneOffset.UTC) > lastSync) {
+                                YouTube.likeVideo(song.id, true)
+                            } else {
+                                database.update(song.song.localToggleLike())
+                            }
                             delay(DB_OPERATION_DELAY_MS)
                         } catch (e: Exception) {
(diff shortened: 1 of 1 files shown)


kairosci:
@suicedia could you verify this solves your problem? Thanks

kairosci:
TESTED: it seems to not overwrite likes and it works as well.

cc: @nyxiereal

nyxiereal (changes requested):
The unconditional union reverses deliberate remote unlikes on every sync. Please add durable reconciliation that distinguishes pending offline local likes from remote removals, then test repeated sync, failure, retry, and process restart.

nyxiereal:
`LastFullSyncKey` is still advanced after `executeSyncLikedSongs()` even when that function catches a reconciliation failure internally. A later sync can then treat an offline like as old and remove it instead of uploading it. Please propagate success/failure and only advance the boundary after successful reconciliation, with tests for API failure, retry, and process restart.

nyxiereal:
Closing as unsafe. The current union-based reconciliation can lose failed offline likes and resurrect intentional remote unlikes. A replacement needs durable pending reconciliation and retry coverage.

PR #3948, opened 2026-06-09
Fix name returning null from Discord's responseBody

brb-fr:
## Problem

When logging in with Discord for the RPC, sometimes the global_name variable won't be available.

## Cause

responseBody not packing global_name with it

## Solution

- If global_name is null use username instead

## Testing

Tested on Honor x7

## Summary by CodeRabbit

* **Refactor**
  * Improved Discord display-name logic to use the global name only when present, otherwise reliably fall back to the username.
* **Bug Fixes**
  * Fixed cases where display names could be empty or incorrect when global name was missing.


nyxiereal:
Test it

brb-fr:
lets see

brb-fr:
> Test it

Screenshot_20260609_211409

no longer says "null"

nyxiereal (approved):
LGTM

PR #3950, opened 2026-06-09
fix(innertube): new player.sj - correct PoToken binding and cipher deobfuscation - add new client visionOS - fix age-restricted/uploaded

alltechdev:
## Problem
Playback fails for age-restricted content and some normal content.

## Cause
1. PoToken binding reversed - playerRequestPoToken received video-bound token instead of session-bound
2. Session ID used dataSyncId for logged-in users instead of visitorData
3. Cipher deobfuscation failed for VM-dispatch players (69e2a55d, 9d2ef9ef) - hardcoded config required Q-array detection
4. WEB_CREATOR missing useWebPoTokens flag

## Solution
- Swap PoToken binding in PoTokenGenerator.kt
- Always use visitorData for session
- Add expression-based cipher configs for 69e2a55d, 9d2ef9ef players
- Remove Q-array check for hardcoded fallback
- Add mutex to CipherDeobfuscator
- Add VISIONOS to fallback clients
- Add stream client display in media info

## Testing
Verified playback for age-restricted songs and normal content with current player hashes.

## Summary by CodeRabbit

* **New Features**
  * Display stream client information in media details.
  * Added support for additional platforms.

* **Improvements**
  * Enhanced playback session handling for consistency.
  * Improved thread-safety for stream deobfuscation.
  * Extended JavaScript-based cipher transform support with prebuilt configurations for improved reliability.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -314,6 +314,7 @@ class MusicService :
     lateinit var connectivityObserver: NetworkConnectivityObserver
     val waitingForNetworkConnection = MutableStateFlow(false)
     private val isNetworkConnected = MutableStateFlow(false)
+    val currentStreamClient = MutableStateFlow<String?>(null)
 
     private lateinit var audioQuality: com.metrolist.music.constants.AudioQuality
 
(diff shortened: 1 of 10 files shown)


alltechdev:
Apparently a side effect of this PR is that uploaded songs are now playing too :)

PR #3857, opened 2026-06-01
[fixed] fix(player): resolve uploaded songs stopping playback after 32s

kairosci:
## Problem
Uploaded songs (privately owned tracks) stop playing after approximately 32 seconds and crash the player when the `contentLength` is null.

## Cause
When streaming URLs are requested for uploaded tracks without providing the `MLPT` playlist context, the YouTube API returns URLs with very short expiration times (around 32 seconds). Additionally, the response formats for these tracks sometimes omit the `contentLength` field, causing a `NullPointerException` when `format.contentLength!!` is executed during the `FormatEntity` upsert.

## Solution
- Pass `playlistId = "MLPT"` to the `playerResponseForPlayback` call in `MusicService` for uploaded songs (detected via a blocking database check on `uploadEntityId`).
- Safely handle missing `contentLength` fields by skipping the `FormatEntity` upsert for these tracks, preventing the crash.

## Testing
Built the app successfully. The data source factory correctly identifies uploaded tracks via the database and modifies the playback response request. 

## Related Issues
- Closes #3710

## Summary by CodeRabbit

* **New Features**
  * Persist optional upload playlist IDs for songs and surface them in media metadata and sync flows.
* **Bug Fixes**
  * More reliable detection of uploaded/private tracks during playback and casting.
  * Safer handling when content length is missing—logs a warning and avoids crashes.
* **Chores**
  * Database migration added to support the new upload playlist field.

PR #3952, opened 2026-06-09
Feat/3604 streaming uploads

### LINKED ISSUE #3604: redesign the upload feature
### Checklist

### Feature description

there's no way i should wait one by one for this to be uploaded to Metrolist. Maybe, add background process for this

Image

less pain

### Additional information

_No response_

luigiwwmf:
## Problem
  The existing "upload to YouTube Music" flow (Library → Songs and the
  "Uploaded" auto-playlist) is unreliable and resource-heavy:

  - **Uploads are killed by navigation/backgrounding.** The loop runs in a 
  Compose `rememberCoroutineScope`, so leaving the screen cancels the whole
  batch.
  - **OOM risk on large files.** Each file is read fully into a `ByteArray`
    (up to ~300 MB) and that same array is handed to Ktor via `setBody(data)` —
    two heap-resident copies per file.
  - **Strictly sequential.** Files upload one-by-one; the next can't start until
    the previous finishes.
  - **No durable retry, no progress UI, no persistence.** A missing
    `X-Goog-Upload-URL` header throws a plain `Exception` that escapes the
    `IOException`-only retry, failing the whole job with just a Toast.
  - **~222 lines of byte-for-byte duplicated upload code** between
    `LibrarySongsScreen` and `AutoPlaylistScreen`.

  ## Cause
  The upload pipeline was implemented as an in-Compose coroutine loop with the
  whole file buffered in memory. Its lifecycle is tied to the composable rather
  than to the work itself, there is no persistent job state, and the retry
  helper
  only wraps a single HTTP request (catching `IOException` only) rather than the
  job as a whole.

  ## Solution
  - **Stream uploads instead of buffering.** `uploadSong`/`uploadSongData` now
  take
    a `contentLength` + `() -> InputStream` factory and stream the body via a
  Ktor
    `ReadChannelContent`; a fresh stream is created per attempt so memory stays
  flat
    regardless of file size.
  - **Persistent upload queue (Room, schema v39).** New `UploadQueueEntity`
    (state/progress/attempts) + DAO, with auto-migration 38→39, as the single
  source
    of truth for job state.
  - **Foreground `UploadService`** owns queue execution so uploads survive
    navigation and process death; orphaned `RUNNING` rows are re-marked
  `PENDING`
    on start.
  - **`UploadViewModel`** drives `PENDING` rows with capped concurrency via a
    `Semaphore` (no longer strictly sequential).
  - **Job-level retry** (`withJobRetry`): exponential backoff, 3 attempts,
  returns
    early on `CancellationException`.
  - **`UploadStatusBar`** — collapsible progress UI; cancel / cancel-all / retry
    wired through the bar and the service notification.
  - **De-duplicated** `LibrarySongsScreen` and `AutoPlaylistScreen` onto the
  shared
    `UploadViewModel` queue.
  - Library auto-syncs (`syncUploadedSongs`) after a run drains with ≥1 success.

  ## Testing
  - Testing was done in Android Studio Pixel 9 with Android 16
  - Queued multiple files from both Library → Songs and the "Uploaded" playlist;
    verified parallel progress in the status bar.
  - Navigated away / backgrounded the app mid-upload — uploads continued via the
    foreground service and the library synced on completion.
  - Verified cancel, cancel-all, and retry from both the status bar and the
    notification.
  - Confirmed large-file uploads complete with flat memory (no `ByteArray`
    buffering).
  - Room migration 38→39 verified against exported schema `39.json`.

  ## Related Issues
  - Closes #3604

## Summary by CodeRabbit

* **New Features**
  * Added background upload queue system for managing file uploads.
  * Added upload status bar displaying progress with cancel, retry, and clear completed actions.
  * Added notification channel for upload operations.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -384,6 +384,7 @@ dependencies {
     implementation(libs.viewmodel)
     implementation(libs.viewmodel.compose)
     implementation(libs.lifecycle.process)
+    implementation(libs.lifecycle.service)
 
     implementation(libs.material3)
     implementation(libs.palette)
--- a/app/schemas/com.metrolist.music.db.InternalDatabase/39.json
+++ b/app/schemas/com.metrolist.music.db.InternalDatabase/39.json
@@ -1,0 +1,1438 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 39,
+    "identityHash": "48901da27585d656863ce22a47e20daa",
+    "entities": [
+      {
+        "tableName": "song",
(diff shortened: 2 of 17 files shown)


nyxiereal:
unresolved issues, high vibecode probability, conflicts with #3857, bumps db version (not allowed)

Issue #3910, opened 2026-06-04
v13.5.0 PLAYBACK/CRASH ISSUES MEGATHREAD


nyxiereal:
# Reporting playback issues in version 13.5.0

We are aware of playback issues in version 13.5.0 and need your help to identify and fix them.

## What we need from affected users

Please provide the following in a comment below:
- **Screenshot** of the issue you're facing
- **App version** (exact version, especially if using a CI/beta build)
- **Debug log** from the app's built-in error reporter (if applicable)
- **Error type** - is it an immediate crash on launch, or just a playback issue
- **Logcat recording** (if you can replicate the issue) - see instructions below

### Firstly, download and install these apps
- [F0x1d/LogFox](https://github.com/F0x1d/LogFox/releases/download/v2.1.2-71/LogFox-v2.1.2-release.apk)
- [thedjchi/Shizuku](https://github.com/thedjchi/Shizuku/releases/download/v13.6.0.r1290-thedjchi/shizuku-v13.6.0.r1290-thedjchi.apk)

### Then follow these instructions
1. Open **Shizuku**, and click on the "View command" button
2. Copy the command and run it on your computer (see ADB setup below if needed)
3. Once Shizuku is running, open **LogFox** and start a recording
4. Open the app and reproduce the bug you're experiencing
5. When finished, press **Export** in LogFox and attach the file to your comment
6. Stop the process by pressing **Exit** in the LogFox notification
7. Stop Shizuku (unless you're using it for other apps)

### Or the video guide
https://github.com/user-attachments/assets/30659a03-39d5-4400-b681-4238c608aebd

> [!IMPORTANT]
> **If you're reporting an immediate crash** issue, you should allow LogFox notification permissions on launch. Then open the app and look for a crash report notification from LogFox. 

> [!TIP]
> If you export the recording as a zip file, upload it to https://litterbox.catbox.moe (set expiration to 3 days).

## Installing ADB on windows

If you don't have ADB installed, follow these steps:

1. Download [Platform Tools](https://dl.google.com/android/repository/platform-tools-latest-windows.zip)
2. Extract the zip file to get the `platform-tools` folder
3. Open Command Prompt in the `platform-tools` folder
4. Test your connection: run `adb.exe devices`
   - Accept the ADB authorization popup on your phone if prompted
5. When executing Shizuku's command, use `adb.exe shell [...]` instead of `adb shell [...]`

---

### Example report
> The app does not let me play music
[...]

PR #3970, opened 2026-06-14
fix(playback): resolve infinite loading and player freeze loops on network errors

### LINKED ISSUE #3958: Issue with 13.5.0
### Checklist - Required

I updated the app to the latest version 

### Expected behavior

The app should have not had play back issues 

### Actual behavior

Had play back issues 

Image 

### Logs

N/a

### Metrolist version

13.5.0

### Android version

ANDROID 16

### Additional information

_No response_

kairosci:
## Problem
Users experienced several severe playback anomalies ("Error code 2000", jumping forward and freezing after track transitions, multiple track skips, and infinite loading loops). This issue was particularly apparent following recent updates and changes to YouTube streaming API.

## Cause
The root cause of the playback freezing and infinite loading was `onPlayerError` improperly capturing network errors (like TCP resets or 403 Forbidden errors due to expired/bad PoTokens/VisitorData) as "No Internet Connection" failures as long as the device had network access. 
When `isNetworkRelatedError()` triggered `waitOnNetworkError()`, the app paused the ExoPlayer and infinitely waited for a network state change, effectively freezing the playback.
Additionally, when generic IO errors did correctly fall through, the recovery path called `player.stop()` instead of just seeking, causing issues with playback state and position.

## Solution
- Removed `isNetworkRelatedError(error)` from the "offline" error branch in `onPlayerError`, ensuring it only waits indefinitely when the device is completely offline (`!isNetworkConnected.value`).
- Routed `isNetworkRelatedError(error)` directly to `handleGenericIOError`, which correctly flushes caching and aggressively re-resolves URLs.
- Replaced `player.stop()` with `player.seekTo` in `handleGenericIOError` to maintain the user's playback position and correctly re-prepare the stream transparently.

## Testing
Simulated simulated TCP resets and API token expiration during playback transitions and verified that the player seamlessly recovers without freezing or resetting the queue position.

## Related Issues
- Closes #3958
- Related to #3910

## Bug Fixes
- Improved playback error recovery by distinguishing offline vs network-related failures: offline now enters a “wait for network” retry flow, while connected network issues use safer generic IO recovery.
- Made recovery after cache clearing more reliable by only resuming/seeking when the current track index is valid.
- Enhanced connectivity monitoring to emit accurate connection state on both network availability changes and capability changes.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -66,7 +66,9 @@ abstract class GenerateProtoTask : DefaultTask() {
             val url = protocUrl.get()
             logger.lifecycle("Downloading protoc ${url.substringAfterLast('/')} from $url")
             protocFile.parentFile.mkdirs()
-            URL(url).openStream().use { input ->
+            val connection = URL(url).openConnection() as java.net.HttpURLConnection
+            connection.setRequestProperty("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36")
+            connection.inputStream.use { input ->
                 protocFile.outputStream().use { output ->
                     input.copyTo(output)
                 }
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -68,6 +68,10 @@ abstract class GenerateProtoTask : DefaultTask() {
             protocFile.parentFile.mkdirs()
             val connection = URL(url).openConnection() as java.net.HttpURLConnection
             connection.setRequestProperty("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36")
+            val responseCode = connection.responseCode
+            if (responseCode !in 200..299) {
+                throw GradleException("Failed to download protoc: Server returned HTTP response code $responseCode for URL: $url")
+            }
             connection.inputStream.use { input ->
                 protocFile.outputStream().use { output ->
                     input.copyTo(output)
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -2892,9 +2892,15 @@ class MusicService :
                 return
             }
 
-            !isNetworkConnected.value || isNetworkRelatedError(error) -> {
-                Timber.tag(TAG).d("Network-related error detected, waiting for connection")
+            !isNetworkConnected.value -> {
+                Timber.tag(TAG).d("No internet connection, waiting for connection")
                 waitOnNetworkError()
+                return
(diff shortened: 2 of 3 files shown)


mounteen0:
[device.txt](https://github.com/user-attachments/files/28928017/device.txt)
[recorded.txt](https://github.com/user-attachments/files/28928018/recorded.txt)
https://litter.catbox.moe/0ydp4ev170sfi485.mp4

The same error appears to occur.
In fact, it’s become even easier to reproduce.
I was able to easily reproduce it by listening to a song over Wi-Fi, closing the app from the app switcher, turning off Wi-Fi, and switching to LTE. It works the other way around, too.
However, when the app is running, there’s no problem even if the network type changes.

kairosci:
@mounteen0 could you retry with [this](https://github.com/MetrolistGroup/Metrolist/actions/runs/27504187957/artifacts/7622779809)? Thanks!

iPrintScreen:
> @mounteen0 could you retry with [this](https://github.com/MetrolistGroup/Metrolist/actions/runs/27504187957/artifacts/7622779809)? Thanks!

This worked for me.

mounteen0:
> @mounteen0 could you retry with [this](https://github.com/MetrolistGroup/Metrolist/actions/runs/27504187957/artifacts/7622779809)? Thanks!

It works as intended.

As noted above, when I try to switch network types after closing the app, the loading process takes longer about once every four attempts; it lasts for about 20 to 30 seconds before completing.

At least during the dozens of attempts I’ve made so far, I haven’t encountered the infinite loading error you reported.

Thank you.

kairosci:
@mounteen0 I think I've fixed the delay too; if you'd like, give it one last test. Thanks so much! The download will appear [here](https://github.com/MetrolistGroup/Metrolist/actions/runs/27535064843)

mounteen0:
> @mounteen0 I think I've fixed the delay too; if you'd like, give it one last test. Thanks so much! The download will appear [here](https://github.com/MetrolistGroup/Metrolist/actions/runs/27535064843)

I tested it using the same method as before and repeated the process 20 times, but I did not observe any loading delays beyond the normal range.
Since this was a manual count rather than an exact measurement, there may be some margin of error, but given that the difference between the longest and shortest times was only about 2–3 seconds, it appears the issue has not resurfaced.
I was initially concerned when the account indicator in the top-right corner of the home screen briefly displayed "Guest," making me wonder if the issue had resurfaced. However, since music playback is functioning normally and I can access my account playlists, it appears that only the visible icons and information on the screen failed to refresh.
While further testing is warranted, based on the current situation, it seems safe to conclude that the issue has been resolved.
Thank you.

kairosci:
cc: @nyxiereal

PR #3978, opened 2026-06-15
Remote-updatable cipher configs + stream-source controls

alltechdev:
## What this does
Ship YouTube player-cipher fixes **without an APK update**, add user-facing
stream-source controls, and harden signature-timestamp resolution.

### Remote-updatable player configs  ← headline
The hardcoded player-config map is replaced by a bundled `player_configs.json`
overlaid by the same file fetched at runtime (6 h TTL + failure-triggered
self-heal). When YouTube rotates `player_ias` and breaks sig/n deciphering,
**publishing the new config fixes every installed device automatically — no
release required.** Payloads are schema-validated and regex-locked before use;
any fetch/parse error falls back to the last-good (or bundled) configs, so
playback is never blocked on the network.

### Stream Sources screen
Toggle which innertube clients are used for stream resolution and view the
effective resolution order. iOS/iPadOS share one underlying client, so they're
a single toggle.

When a client is completely broken, you can toggle it off to avoid trying on every single song, and instead use what works until a fix is pushed.

### Reliability
- Prefer the cipher player's own signature timestamp to avoid CDN 403s when
  YouTube A/B-rolls a different player generation; cooperative cancellation
  across the decipher path.

### Song details
Show the player hash and the date cipher support was added.

> Remote configs (and the cosmetic "cipher support added" dates) are fetched
> from `ZemerTeam/zemer-cipher`. Offline/failure → bundled asset only.

### Screenshots

Screenshot_20260615-134817
Screenshot_20260615-134838_com metrolist music debug

## Summary by CodeRabbit

* **New Features**
  * Added a “Stream sources” settings section to enable/disable stream clients and control their priority order.
  * Added additional media info details for web stream formatting (e.g., player hash and cipher support status).
* **Bug Fixes**
  * Playback now avoids attempting disabled stream clients and improves stream fallback ordering.
  * Improved handling of in-flight request cancellations to prevent masking errors.
* **UI / Content**
  * Added new strings and an icon for stream-source display and formatting.

### DIFF
--- a/app/src/main/assets/player_configs.json
+++ b/app/src/main/assets/player_configs.json
@@ -1,0 +1,19 @@
+{
+  "schemaVersion": 1,
+  "players": {
+    "69e2a55d": { "sig": "Jf(20,3699,INPUT)", "nClass": "iE", "sts": 20611, "aliases": ["70d8066f"] },
+    "9d2ef9ef": { "sig": "v0(35,4499,INPUT)", "nClass": "uY", "sts": 20607, "aliases": ["6fb43da5"] },
+    "16ee6936": { "sig": "mP(4,155,INPUT)", "nClass": "Yx", "sts": 20613, "aliases": ["ca366632"] },
+    "ce74690f": { "sig": "$9(2,6487,INPUT)", "nClass": "cV", "sts": 20612, "aliases": ["a5669e32"] },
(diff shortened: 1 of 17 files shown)


alltechdev:
Mo asked me for this. The only thing not in Metrolist repo is the jsons for configs and player dates. That's on my cipher repo, https://github.com/ZemerTeam/zemer-cipher. When I update it, ML users will receive the update automatically. No more apk updates for youtube player changes, unless they do a big change for extraction.

Been using this on an ML fork - https://github.com/ZemerTeam/zemer-app, and it's been real solid.

PR #3998, opened 2026-06-19
fix(playback): eliminate PlayerConnection NPE race with MusicService player init

### LINKED ISSUE #3943: The whole app doesn't load anything
### Checklist - Required

The bug happens everytime I restart my phone, and the solution is to sign out and then sign back in. This bug has never occured in the older versions of Metrolist.

### Expected behavior

The app should load just fine, displaying reccomended songs, revisited songs, etc.

### Actual behavior

The actual behavior is shown in the video below.

https://github.com/user-attachments/assets/e1013eca-94f4-4df1-af46-0d44327156a0

### Logs

[07_06-20-41-14_865.log](https://github.com/user-attachments/files/28681187/07_06-20-41-14_865.log)

### Metrolist version

13.5.0

### Android version

Android 15

### Additional information

_No response_

kairosci:
## Summary
- Fix NPE crash in `PlayerConnection` construction caused by a race condition between `onServiceConnected` and `MusicService.player` (lateinit var) initialization
- `PlayerConnection` now always constructs successfully with safe defaults; state syncs automatically via `playerFlow` when the player becomes ready

## Root cause
When `MusicService.onCreate()` is still running and `onServiceConnected` fires, `PlayerConnection` was accessing `service.player` before it was initialized. With R8, the `lateinit` check is stripped, so `service.player` returns null silently — any subsequent method call (e.g. `.playbackState`) throws NPE.

Three unprotected `player` accesses during `PlayerConnection` construction (lines 146, 149, 208 in the old code) bypassed the try/catch in the first `init` block, crashing the entire construction. The retry in `MainActivity.onServiceConnected` also failed because the player was still not ready after 500ms.

## Fix (structural)
1. **`getPlayerSafe()`** — Now calls `check(playerReadinessFlow.value)` before touching `service.player`, preventing the NPE at the source
2. **`getPlayerOrNull()`** — New safe accessor used during construction; returns `null` when player isn't ready, catching both `UninitializedPropertyAccessException` and `NullPointerException`
3. **`initialState`** — Computed via `getPlayerOrNull()` before any property init, providing safe defaults for `isEffectivelyPlaying` and `mediaMetadata`
4. **`isEffectivelyPlaying`** — Uses `initialState.third` instead of `player.playbackState` (was the primary NPE source)
5. **`mediaMetadata`** — Uses `getPlayerOrNull()?.currentMetadata` instead of `player.currentMetadata`
6. **Second `init` block** — Removed try/catch-rethrow; uses `getPlayerOrNull()` for immediate attachment
7. **`MainActivity.onServiceConnected`** — Removed retry logic since construction can no longer throw

## Testing
Verified that all state flows initialize with safe defaults and that `playerFlow` collector registers correctly to sync real state when MusicService finishes player initialization.

Fixes #3943

## Summary by CodeRabbit

* **Refactor**
  * Simplified player connection initialization and state management.
  * Updated player access patterns to use safer, readiness-aware accessors.
  * Streamlined connection setup logic by removing retry behavior.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/MainActivity.kt
+++ b/app/src/main/kotlin/com/metrolist/music/MainActivity.kt
@@ -204,7 +204,6 @@ import com.metrolist.music.widget.PlaylistWidgetReceiver
 import com.valentinilk.shimmer.LocalShimmerTheme
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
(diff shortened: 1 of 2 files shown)


kairosci:
@3kstra could verify and confirm if this solves your problem? Thanks!
Click [here](https://github.com/MetrolistGroup/Metrolist/actions/runs/27818659550/artifacts/7747113197) to download app

3kstra:
Sorry for the late reply @kairosci, super busy.

At first glance, the app seems to work fine. I logged in and then did a restart and quick a shutdown-startup, and the app started just fine.

I assume the app runs fine now, but I'll see tomorrow.

kairosci:
No problem, I'll wait for your confirmation to merge it.

3kstra:
Awesome, it works all good now, even compared it to 13.5.0 and can confidently say the debug fixed the issue and runs fine. Thank you so much @kairosci, Godspeed to you and Metrolist!

kairosci:
cc @mostafaalagamy

PR #4015, opened 2026-06-21
fix(playback): handle ERROR_CODE_REMOTE_ERROR with stream URL refresh instead of generic retry

kairosci:
## Problem
ExoPlayer ERROR_CODE_REMOTE_ERROR silently stops playback without retry, leaving the user on a stuck track.

## Cause
The onPlayerError handler did not recognize REMOTE_ERROR codes and fell through to stopOnError. Additionally, the generic retry handleGenericIOError does not invalidate the cached stream URL, so retrying with an expired URL fails again.

## Solution
Added isRemotePlaybackError check in onPlayerError that calls handleExpiredUrlError, which clears the cached stream URL, marks WEB_REMIX as failed, refreshes cipher config, and retries playback. This gives a structural recovery instead of a blind retry.

## Testing
Tested on emulator with network interruptions triggering remote errors, verified retry fires with fresh URL resolution.

## Bug Fixes
* Improved error handling and recovery mechanisms for remote playback connection failures, allowing the app to gracefully manage interrupted streams
* Enhanced detection and user-facing error messaging when videos become unavailable for playback, with clear guidance to retry or skip to the next song

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MusicService.kt
@@ -2970,6 +2970,9 @@ class MusicService :
         error.errorCode == PlaybackException.ERROR_CODE_IO_FILE_NOT_FOUND ||
             (error.cause as? PlaybackException)?.errorCode == PlaybackException.ERROR_CODE_IO_FILE_NOT_FOUND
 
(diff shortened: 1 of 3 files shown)


kairosci:
With this it seems to work again !

---



Before this in 13.6.0.

PR #4080, opened 2026-07-01
[wontfix] fix(playback): replace WEB_REMIX with ANDROID_VR_1_43_32 as primary stream source

### LINKED ISSUE #4065: Stream Resource is not working properly.
### Checklist - Required

1. Open app
2. Play music
3. And music not loading

### Expected behavior

In the Stream Resources section WEB_REMİX The source is delaying the loading and playback of the music.When I disable it, it plays songs very quickly. I've added a few photos to the topic, Second photo shows the problem WEB_REMİX when it is open.

### Actual behavior

The "WEB_REMIX" resource should be corrected or removed.

Image
Image

### Logs

N/A

### Metrolist version

13.6.0

### Android version

Android 16 Hyperos 3 

### Additional information

_No response_

kairosci:
WEB_REMIX is used as the primary stream client but requires computationally expensive web-based transformations for every stream URL including signature timestamps, WebView-based poToken generation, and cipher-based n-transform deobfuscation, causing noticeable playback delays. When users disable WEB_REMIX in Settings the fallback clients stream immediately, confirming WEB_REMIX is the bottleneck.

The root cause is that WEB_REMIX was set as MAIN_CLIENT in YTPlayerUtils, making it the first client attempted for every stream resolution. When WEB_REMIX provides an OK response the code skips HEAD validation and exits the fallback loop immediately, so the lightweight native clients (VISIONOS, ANDROID_VR) are never tried.

MAIN_CLIENT is now ANDROID_VR_1_43_32 which returns direct CDN audio URLs without requiring any signature timestamp, poToken, or n-transform operations. The stream URL is usable immediately from the player response with standard HEAD validation. WEB_REMIX is moved to the end of the fallback chain as a last resort when all other clients fail. The playerResponseForMetadata function continues using WEB_REMIX for rich metadata retrieval as it already has its own separate code path.

Since the upfront PoToken generation was tied to MAIN_CLIENT.useWebPoTokens and MAIN_CLIENT is now ANDROID_VR_1_43_32 which does not use web tokens, the PoToken generation is made lazy and triggered only when a fallback client with useWebPoTokens (such as WEB_REMIX at the end of the chain) is reached. This ensures the WEB_REMIX fallback still appends pot= to the streaming URL without imposing the WebView startup cost on the fast path.

Closes #4065.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback reliability by refining how streaming clients are selected and how fallback is handled.
  * Stream URL validation is now applied more consistently, reducing cases where videos fail to load or play.
  * Improved metadata retrieval by switching to a more suitable request path, which may enhance content details and age-restricted playback behavior.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
@@ -162,7 +162,7 @@ object YTPlayerUtils {
         val signatureTimestamp = getSignatureTimestampOrNull(videoId)
         Timber.tag(logTag).d("Signature timestamp: ${signatureTimestamp.timestamp}")
 
-        val poToken: PoTokenResult? = null
+        var poToken: PoTokenResult? = null
 
         // Try MAIN_CLIENT
         Timber.tag(logTag).d("Attempting to get player response using MAIN_CLIENT: ${MAIN_CLIENT.clientName}")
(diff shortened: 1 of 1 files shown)


nyxiereal:
thats not what we want

PR #4090, opened 2026-07-02
feat(integrations): import Spotify library via Exportify CSV

selmansem:
## Problem
Users coming from Spotify have no in-app way to bring their library into Metrolist. The existing CSV import works but is generic and requires manual column mapping, so it isn't discoverable or approachable for that use case.

## Cause
A first-party Spotify sync isn't viable: as of Feb 2026, Spotify's Development Mode requires the app owner to hold Premium and caps usage at a handful of allowlisted users, and extended quota requires a registered business with 250k+ MAU. So Metrolist can't ship a direct Spotify Web API integration that serves all users.

## Solution
- Add a "Spotify" entry under Settings → Integrations.
- Guide the user to export their library via Exportify (which runs in Spotify's Extended Quota Mode and is exempt from the Development Mode limits — no Premium, no credentials shipped in the app).
- Auto-detect Exportify's CSV column layout (`Track Name`, `Artist Name(s)`) so the manual column-mapping step is skipped.
- Reuse the existing CSV matcher to resolve tracks against YouTube Music.
- Add a `spotify.xml` drawable for the integration icon.

## Testing
Built `:app:assembleFossDebug` and installed on a physical device (Redmi Note 9 Pro). Imported a real playlist end to end and confirmed tracks resolve into a Metrolist playlist. Also verified the unrecognized-CSV path shows the friendly error message instead of failing silently.

## Summary by CodeRabbit

* **New Features**
  * Added Spotify integration settings with an import flow that lets users pick an Exportify-generated CSV and automatically imports the data.
  * Added a Spotify entry to the Integrations screen, including a new Spotify icon.
  * Included an import progress experience and a dedicated screen with navigation controls.
* **Bug Fixes**
  * Improved CSV import reliability with automatic detection of title/artist columns.
  * Added clearer user messaging for empty or unrecognized CSV files.


selmansem:
Heads-up on one design decision: the `spotify.xml` drawable uses the recognizable Spotify glyph to mark the integration. I went with it because it makes the entry instantly identifiable in the integrations list, and it's tinted with the theme color (`?attr/colorControlNormal`) rather than Spotify green, to stay consistent with the other integration icons.

That said, it's a trademarked mark, so I'm happy to swap it for a generic icon (e.g. a neutral import/library glyph) if you'd prefer to avoid shipping the brand asset. Just let me know which way you'd like to go.

selmansem:
Addressed the actionable items:

- Progress/log callbacks are now wired into the loading screen (state changes dispatched on the main thread).
- Real import failures are handled separately from empty results.
- The dialog title uses a string resource instead of a hardcoded literal.
- KDoc and inline comments translated to English.

On the CSV tokenizer duplication: I've left `parseCsvHeader` self-contained for now, since reusing `BackupRestoreViewModel.parseCsvLine` would mean widening a currently-private helper in a shared ViewModel. Happy to change that if a maintainer prefers.

selmansem:
**Known limitations**

Both are pre-existing in the shared `AddToPlaylistDialogOnline` matcher used by all import methods (CSV, M3U, and this Exportify flow), so they're intentionally left out of this PR:

- Imports don't dedupe against songs already present, so re-importing an overlapping playlist can create duplicates.
- With no network connection, the import finishes with 0 results and no error, because each per-track YouTube Music lookup fails silently in the matcher.

I understand the project is in maintenance mode and not accepting new features right now — no rush at all on this. Leaving it open in case it's useful whenever that changes, and happy to rebase or adjust if needed.

nyxiereal:
well, as previously discussed in multiple other releases, we are currently in maintenance mode, no new features will be added to the application as we're currently rewriting the app in kotlin multiplatform.

other than that, importing songs from spotify is planned (but only by me lol) in the rewrite.

to avoid confusion, i will put a warning in the repo itself

PR #4092, opened 2026-07-03
fix(player): recover from WebView renderer death; surface swallowed stream-resolution failures

alltechdev:
## Problem
On low-RAM devices under sustained memory pressure, playback can break for the rest of the session: every song stalls for ~12–22 seconds and then falls back to the unreliable non-WebView paths (or fails outright), until the app is restarted. Separately, when stream resolution fails, the logs don't show why — a network/parse error and a genuine "not playable" rejection both surface as `Status NOT OK: null`, making these failures very hard to diagnose from user reports.

## Cause
- The kernel can OOM-kill the sandboxed WebView renderer process. The cipher WebView (and the poToken WebView) never noticed: the dead instance stayed cached for the whole session, so every decipher call hung until its 15 s timeout. On old WebView providers `onRenderProcessGone` is never delivered at all.
- The per-client `player()` calls used bare `.getOrNull()`, silently discarding thrown request/deserialization failures.

## Solution
- Handle `onRenderProcessGone` in `CipherWebView` and `PoTokenWebView`: fail all pending continuations immediately (fail fast instead of waiting out the 15 s timeout), destroy the dead instance, and rebuild it on demand for the next call.
- Add evaluate/init timeout safety nets that treat a non-responsive renderer as gone, covering old WebView providers that never deliver the callback.
- Add `RendererRecoveryPolicy`: after 3 consecutive renderer deaths, skip the doomed multi-second player.js re-parse for a short (60 s, half-open) backoff window, then retry — the WebView stays the primary path.
- Treat poToken renderer death as transient instead of a `BadWebViewException` that permanently disables poTokens for the session.
- Log per-client `player()` request failures before `getOrNull()` so thrown failures are distinguishable from playability rejections.

## Testing
`assembleFossDebug` built and tested on-device (OnePlus 12, Android 16, rooted). Renderer failures were forced with `su -c kill` on the app's sandboxed WebView renderer process — SIGKILL is the same signal the kernel OOM killer delivers, and SIGSTOP simulates a wedged renderer where `onRenderProcessGone` is never delivered. Each signal was verified via `/proc/<pid>` state before counting a trial.

**Renderer killed (OOM simulation), 5/5 recoveries:**

| | detection (kill → `onRenderProcessGone`) | rebuild → decipher success | skip → playback data |
|---|---|---|---|
| range over 5 runs | 50–73 ms | 184–201 ms | 1.6–3.3 s |

(baseline skip → playback with a healthy renderer: ~0.6 s; the renderer PID changed on every iteration, confirming a genuinely new process each rebuild)

**Wedged renderer (SIGSTOP, callback never fires):** all timeout safety nets fired — sig eval 15 s and WebView-init 30 s both ended in "treating renderer as gone"; poToken attempts were capped at 8 s and playback proceeded without a poToken. Every affected song still played via fallback clients ~1 s after the cipher gave up; zero playback failures across the whole test.

**Backoff (`RendererRecoveryPolicy`):** after 3 consecutive renderer deaths the next song logged "Skipping cipher WebView creation: 3 consecutive renderer deaths, in backoff window" and skipped the doomed ~30 s rebuild instantly; after the 60 s window expired, a fresh WebView was created and deciphered in ~150 ms with the policy fully reset.

For comparison, before this fix a single renderer death left the cached WebView a session-permanent zombie: every subsequent song stalled 12–22 s until app restart.

## Related Issues
- None

## Summary by CodeRabbit

* **New Features**
  * Added smarter recovery for WebView-based playback and token generation after renderer crashes or timeouts, including controlled retry backoff.

* **Bug Fixes**
  * Improved fallback behavior when playback/resolution requests fail, ensuring continued playback attempts instead of stalling.
  * Hardened WebView lifecycle handling for safer initialization/teardown, preventing hangs/crashes and improving resilience during renderer death.
  * Token generation now proactively refreshes when the existing WebView session is marked dead.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/YTPlayerUtils.kt
@@ -202,7 +202,12 @@ object YTPlayerUtils {
             // Age-restricted: use WEB_CREATOR directly (no NewPipe needed from here)
             Timber.tag(logTag).d("Age-restricted detected, using WEB_CREATOR")
             Timber.tag(TAG).i("Age-restricted: using WEB_CREATOR for videoId=$videoId")
-            val creatorResponse = YouTube.player(videoId, playlistId, WEB_CREATOR, null, null).getOrNull()
+            val creatorResponse = YouTube.player(videoId, playlistId, WEB_CREATOR, null, null)
+                .onFailure {
+                    // Distinguish thrown request/parse failures from genuine playability
+                    // rejections (both otherwise surface as a null response downstream).
+                    Timber.tag(logTag).e(it, "player() request FAILED for WEB_CREATOR")
+                }.getOrNull()
             if (creatorResponse?.playabilityStatus?.status == "OK") {
                 Timber.tag(logTag).d("WEB_CREATOR works for age-restricted content")
                 mainPlayerResponse = creatorResponse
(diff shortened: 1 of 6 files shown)


alltechdev:
If ya merge, please make sure coauthor @jask625 gets credit, he is the one who found the bug on his 2gb ram device.

nyxiereal:
@alltechdev ready to merge when you are :saluting_face:

alltechdev:
> @alltechdev ready to merge when you are 🫡

Gimme a sec I'll shut up coderabbit and run one more test batch

alltechdev:
All good, ready to go.

PR #4097, opened 2026-07-05
fix(playlists): sync playlist removals from menus

yeongjo:
## Problem
The `Remove from playlist` action was missing from the player menu when
selecting a currently playing song from a playlist.

## Cause
The player menu could not reliably identify the playlist context for the
currently playing item. Playlist browse IDs and set video IDs were
available in some queue paths, but the local playlist ID was not preserved,

## Solution
- Preserve the local playlist ID in list queues.
- Pass local playlist IDs when starting playback from local playlists.
- Use queue playlist context to determine when `Remove from playlist`
should be shown in the player menu.
- Schedule playlist removals through sync utilities instead of relying on
direct removal calls from the menu.
- Remove playlist request metadata handling from YouTube playlist queue
items.
- Keep the YouTube song menu remove action limited to editable playlists.

## Review follow-up
Addressing the request for a rebase and regression coverage on this
destructive path:

- Rebased onto current `main`; the two merge commits are gone and the
  branch is now a clean series on top of `chore(deps): bump InnerTubeX to 0.4.1`.
- Fixed a build break that the rebase surfaced: `adoptQueue()` (added
  upstream in #4214) assigned to `currentQueue`, which this PR turns into
  a read-only `StateFlow`. It now writes `_currentQueue.value`.
- Moved the logic that decides *which* row to delete out of composables
  and out of the fire-and-forget sync coroutine into pure functions, so it
  can be tested directly:
  - `resolvePlaylistRemovalTarget()` / `selectLocalPlaylistRowToRemove()`
    (extracted from `PlayerMenu`)
  - `resolveSetVideoIdForRemoval()` (extracted from `SyncUtils`), which
    makes the remote lookup budget injectable and pins the rule that a
    removal is **abandoned, never guessed**, when no setVideoId resolves.
- Deduplicated the three playlist queue builders in `LocalPlaylistScreen`
  into `Playlist.toListQueue()`, and dropped the `SongItem.toPlaylistItem()`
  indirection that just called `toMediaItem()`.

### Regression coverage
23 new unit tests, covering each case raised in review:

| Case | Tests |
| --- | --- |
| Duplicate song occurrences | `duplicate occurrences of one song resolve to different targets`, `local row lookup picks the occurrence matching the set video id` |
| setVideoId identity | `missing or blank set video id yields no target`, `local row lookup ignores the same song in other playlists`, `local row lookup returns null when no occurrence matches` |
| Restored queues | `QueuePlaylistContextTest` (persist/restore round trip, Java serialization round trip, per-occurrence targets after restore) |
| Non-editable playlists | `non editable youtube playlist yields no target`, `non editable list queue yields no target`, `list queue without browse id yields no target` |
| Remote removal failure / retry | `SetVideoIdResolverTest` (retry until success, retry after a thrown lookup, give up after the attempt budget instead of removing an unknown occurrence) |

### Known migration cost
`PersistQueue` gains three fields, which changes its Java
`serialVersionUID`. A queue persisted by an older build therefore fails to
deserialize once on upgrade. The failure is already caught and the file is
cleared, so there is no crash — users lose the restored queue a single
time, and it is repopulated on the next playback. Flagging it rather than
pinning the previous UID; happy to do that instead if you would prefer to
keep old queues readable.

## Testing
- `./gradlew :app:assembleFossDebug :app:testFossDebugUnitTest` — build successful,
  new tests pass (23 tests, 0 failures, 0 errors).
- Tested on Galaxy S23 (Android 16, One UI 8.5).
- Verified the 'Remove from playlist' action across multiple playlists, cross-checking with YouTube to ensure consistent state sync.

## Related Issues
- Closes #
- Related to #

## Summary by CodeRabbit

* **New Features**
  * Added support for removing songs from editable playlists during playback.
  * Added “Remove from playlist” actions to song and player menus.
  * Preserved playlist details and song mappings during playback and queue restoration.

* **Bug Fixes**
  * Playlist views now accurately reflect removed songs.
  * Improved handling of songs with matching identifiers and playlist mappings.

* **Chores**
  * Improved queue and playlist metadata handling across playback screens.
  * Added more reliable queue persistence and restoration.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/extensions/QueueExt.kt
+++ b/app/src/main/kotlin/com/metrolist/music/extensions/QueueExt.kt
@@ -13,6 +13,7 @@ import com.metrolist.music.playback.queues.ListQueue
 import com.metrolist.music.playback.queues.LocalAlbumRadio
 import com.metrolist.music.playback.queues.Queue
 import com.metrolist.music.playback.queues.YouTubeAlbumRadio
+import com.metrolist.music.playback.queues.YouTubePlaylistQueue
 import com.metrolist.music.playback.queues.YouTubeQueue
 
 fun Queue.toPersistQueue(
@@ -27,7 +28,19 @@ fun Queue.toPersistQueue(
             items = items,
             mediaItemIndex = mediaItemIndex,
             position = position,
-            queueType = QueueType.LIST
+            queueType = QueueType.LIST,
+            playlistBrowseId = playlistBrowseId,
+            playlistId = playlistId,
+            playlistIsEditable = playlistIsEditable,
+        )
+        is YouTubePlaylistQueue -> PersistQueue(
+            title = title,
+            items = items,
+            mediaItemIndex = mediaItemIndex,
+            position = position,
+            queueType = QueueType.LIST,
+            playlistBrowseId = playlistId,
+            playlistIsEditable = isEditable,
         )
         is YouTubeQueue -> {
             // Since endpoint is private, we'll store a simplified version
@@ -84,7 +97,10 @@ fun PersistQueue.toQueue(): Queue {
             title = title,
             items = items.map { it.toMediaItem() },
             startIndex = mediaItemIndex,
-            position = position
+            position = position,
+            playlistBrowseId = playlistBrowseId,
+            playlistId = playlistId,
+            playlistIsEditable = playlistIsEditable,
         )
         is QueueType.YOUTUBE -> {
             // For now, fallback to ListQueue since we can't reconstruct YouTubeQueue properly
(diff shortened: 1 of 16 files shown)


nyxiereal (changes requested):
Song ID maps collapse duplicate playlist occurrences and can remove the wrong item. Please carry occurrence identity, gate removal on playlist editability, publish metadata safely across threads, and test duplicate songs.

nyxiereal:
The final occurrence/editability handling looks improved, but this is a destructive and failure-sensitive path without a current build or regression coverage. Please rebase and add tests for duplicate song occurrences, `setVideoId` identity, restored queues, non-editable playlists, and remote removal failure/retry before merge.

nyxiereal:
@codex review

PR #4103, opened 2026-07-07
Harden cipher/poToken pipeline: token races, WebView leaks, clock-skew wedges, main-thread jank

alltechdev:
## Problem
Several latent defects in the cipher/poToken pipeline degrade or break playback in ways that are hard to diagnose from the field:

- A single transient JS error in Google's remotely-served botguard code **permanently disables poTokens for the whole session** (streams degrade to the ~1 MiB cutoff / 403 pattern until app restart).
- Concurrent poToken mints (playback racing a prefetch/download) can deliver one video's token to another caller, or orphan a caller into a 15 s timeout that then destroys a perfectly healthy WebView.
- A failed session-pot mint leaves poisoned state behind: `KotlinNullPointerException` per song, or silently wrong-session tokens that 403.
- Skipping a song mid-decipher can deliver the **previous song's signature to the next request** (sporadic unexplainable 403s).
- Every cipher WebView build runs a ~2.8 MB string transform + disk write **on the main thread** — visible jank/ANR risk exactly at song start on slow devices.
- Failed WebView inits leak a live renderer process per retry; a backward clock step (NTP correction) wedges the player-config self-heal for the entire skew; process death mid-cache-write can serve a truncated player.js for 6 hours.

## Cause
- Error classification maps any console message containing "Uncaught" (and any "SyntaxError") to `BadWebViewException`, which sets a never-reset `webViewBadImpl` kill switch — even post-init, where such errors come from Google's own botguard program, not a broken WebView engine.
- The injected mint JS used bare page globals (`identifier`), read at promise-resolution time, and pending continuations were keyed by raw identifier (silent overwrite on same-key concurrency).
- `PoTokenGenerator` committed `sessionId`/generator state *before* the fallible session-pot mint.
- The sig/n JS bridge had no pairing between requests and results, so a late callback resumed whatever continuation was currently armed.
- `CipherWebView.create()` only destroyed the half-built WebView on timeout/cancellation, not on error resumes; `loadPlayerJsFromFile` ran entirely inside `withContext(Dispatchers.Main)`.
- Cooldown/TTL checks used signed wall-clock deltas (`now - stamp < window`), which stay true for negative deltas; cache writes were plain truncating `writeText` with no serialization against the purge in a concurrent writer.

## Solution
- `PoTokenWebView`: pending mints keyed by a **per-call unique request key**, captured lexically in an IIFE and echoed back by the JS bridge; post-init "Uncaught" and mint-time JS errors are **transient** (`PoTokenException` + recreate on next call) instead of the permanent `BadWebViewException` kill switch (init-phase broken-WebView detection unchanged); OkHttp responses closed on non-200 paths.
- `PoTokenGenerator`: session state committed **only after** the streaming-pot mint succeeds; the freshly created WebView is closed (not leaked) on mint failure.
- `CipherWebView`: sig/n requests carry an id validated on take (`RequestSlot`) so stale results are ignored; the ~2.8 MB player.js transform + write moved to `Dispatchers.IO` (only WebView construction/load stays on Main); the player JS string is no longer retained for the WebView's lifetime (~2.8 MB heap); the half-initialized WebView is destroyed on **any** init failure.
- `PlayerConfigStore`/`PlayerJsFetcher`: all wall-clock window checks go through one `withinWindow` helper that treats backward clock steps as expired/not-armed; player-JS cache writes are atomic (temp + rename, with delete-and-retry before the non-atomic last resort) and serialized behind a lock.
- `FunctionNameExtractor`: `signatureTimestamp` precedence is now anchored-JS-literal → config → loose pattern, so a bad `sts` in a config update can never shadow the player's own embedded value.
- `PoTokenWebView`: an empty botguard `200` response body is now treated as an init failure via the existing `PoTokenException` path — it previously fell through to `handleResponseBody("")` and failed deeper with a murkier error; the body is also read null-safely.

No changes to the remote config schema, parser, or fetch protocol — existing installs keep fetching configs unchanged.

These fixes were developed, reviewed, and hardened in [zemer-cipher](https://github.com/ZemerTeam/zemer-cipher) (a standalone library extracted from this same cipher/poToken lineage — see ZemerTeam/zemer-cipher#1 for the fully annotated changeset), then carefully ported here with each file diffed against this tree first and adapted to its local structure rather than copied blind.

## Testing
- Cipher unit tests pass (`:app:testFossDebugUnitTest --tests "com.metrolist.music.utils.cipher.*"`); the pre-existing `YouTubeUtilsTest` failures on `main` are unrelated and unchanged.
- On-device (foss debug build, real playback across two app sessions): 26 deciphered stream URLs, 13 n-transform successes, 15 uniquely-keyed poToken mints — every JS bridge callback id-matched, one WebView build per process with the 2.8 MB prep measured at ~50 ms on the IO thread, and **zero** crashes, renderer-gone events, Uncaught/BadWebView classifications, or timeouts in 20k+ captured log lines.
- The same changeset was additionally validated end-to-end against the live CDN in the zemer-cipher repo (real signatureCipher deciphered, n-probe changed, Range GET returned HTTP 206).

## Related Issues
- Related to #4092 (WebView renderer-death recovery — this extends that hardening to the token/decipher race and leak classes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading playback-related data, reducing failures caused by concurrent requests, stale cache entries, and interrupted refreshes.
  * Fixed issues where token generation could return the wrong result or fail during overlapping requests.
  * Made background loading and cache updates more resilient, helping prevent broken or partially written data from being used.
  * Improved handling of unexpected WebView errors so recovery is more consistent.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/cipher/CipherDeobfuscator.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/cipher/CipherDeobfuscator.kt
@@ -80,7 +80,7 @@ object CipherDeobfuscator {
             Timber.tag(TAG).w("signatureTimestamp: could not fetch player JS")
             return null
         }
-        val sts = FunctionNameExtractor.extractSignatureTimestamp(playerJs)
+        val sts = FunctionNameExtractor.extractSignatureTimestamp(playerJs, hash)
         Timber.tag(TAG).d("Cipher player STS (hash=$hash): $sts")
         return sts
     }
--- a/app/src/main/kotlin/com/metrolist/music/utils/cipher/CipherWebView.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/cipher/CipherWebView.kt
@@ -20,15 +20,8 @@ import kotlin.coroutines.Continuation
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 
-/**
- * WebView-based cipher executor for YouTube stream URL deobfuscation
- *
- * Executes signature decipher and n-transform functions extracted from player.js.
- * Supports both regex-extracted functions and hardcoded fallback for Q-array obfuscated players.
- */
 class CipherWebView private constructor(
     context: Context,
-    private val playerJs: String,
     private val sigInfo: FunctionNameExtractor.SigFunctionInfo?,
     private val nFuncInfo: FunctionNameExtractor.NFunctionInfo?,
     initContinuation: Continuation<CipherWebView>,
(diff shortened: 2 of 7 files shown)


nyxiereal (approved):
LGTM

Issue #3210, opened 2026-03-10
Switches songs way to early when Crossfade is enabled and slower speed was set


ImPro396:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

1. Enable crossfade, duration doesn't matter
2. Start a song and under advanced select a slower speed e.g. 0.8
3. Song transitions way too early, the end of the first song gets cut off

### Expected behavior

Song plays to the end, then crossfade transitions to the next song

### Actual behavior

Song gets cut off too early, the slower the speed the earlier

### Logs

Shizuku does'nt work on my phone after connected to my pc, hope it's not required in this case

### Metrolist version

13.3.0

### Android version

14

g4mar1nho:
Hi! I was able to reproduce this bug and would like to work on this issue.

Dr-Brixx:
I was also able to reproduce this.

[04_06-19-38-10_119.txt](https://github.com/user-attachments/files/28607300/04_06-19-38-10_119.txt)

PR #4112, opened 2026-07-09
fix(playback): account for playback speed in crossfade delay calculation to prevent premature track transition

kairosci:
## Problem

When crossfade is enabled and playback speed differs from 1.0x, songs transition to the next track well before their natural end, cutting off the finale of each track. The lower the speed, the earlier the premature transition.

## Cause

The crossfade trigger delay in scheduleCrossfade is computed in media time as player.duration minus crossfadeDuration minus player.currentPosition but kotlinx.coroutines.delay operates in real wall clock time. When playback speed is not 1.0x, media time and real time diverge, causing the crossfade to fire at the wrong media position. For example at 0.8x speed with a three minute song, the crossfade triggers approximately thirty five seconds early instead of the intended five seconds before the end.

## Solution

Divide the media time remaining by the current playback speed to convert it to the correct real time delay before passing it to delay. This ensures the crossfade triggers at the intended media position regardless of speed. Additionally copy the primary players playback parameters to the secondary ExoPlayer created during crossfade so the next track maintains the same speed throughout the transition.

## Testing

Enable crossfade in settings and play songs at various playback speeds both below and above 1.0x. Verify the crossfade transitions consistently start at crossfadeDuration milliseconds before the natural end of each track regardless of the speed setting. Verify the next track maintains the selected speed throughout the crossfade period.

## Related Issues

Related to: https://github.com/MetrolistGroup/Metrolist/issues/3210

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crossfade timing when playback speed changes by scaling the crossfade window with the current speed and triggering the transition based on remaining media time.
  * Improved crossfade reliability by ensuring the fading track uses matching playback parameters before starting secondary playback.


kairosci:
@Dr-Brixx Could you test it and let me know if it solves the problem? Thanks.
I did a quick test, and it seems to work fine.

Dr-Brixx:
If the speed is over 1, the next song plays double during the crossfade

kairosci:
@Dr-Brixx could you take another looks? Thanks

Dr-Brixx:
Isn't working sadly

https://github.com/user-attachments/assets/8a9a45b2-42ef-4fe8-afdb-6e338f0cda0c

kairosci:
@Dr-Brixx as soon as you can, could you take another look? I tested it on several songs, and it doesn't seem like issue is there anymore

Dr-Brixx:
Now crossfade doesn't work at all and I get this error
Screenshot_20260719-235356

kairosci:
It didn't show up on my phone, but in any case, I should have covered this scenario as well (I'm just saying this in theory, since the error doesn't appear on my phone). Could you try again? Thanks

Dr-Brixx:
Now crossfade starts too early again (I set it to 10s but it uses 20s) and it ignores audio normalisation. Also the first song keeps playing even though it switched to the second.


https://github.com/user-attachments/assets/a521e63c-2c19-4488-aa98-69cb1f314e3b

Here are logs (with different songs, but still the same)
[21_07-23-09-10_377.txt](https://github.com/user-attachments/files/30243938/21_07-23-09-10_377.txt)

kairosci:
Tested the latest commit: the songs fade out at the end, both with and without normalization.
The ending is no longer trimmed; I think it's good.

Dr-Brixx:
Looks good to me, I couldn't find any issues. Thank you very much for fixing this

PR #4139, opened 2026-07-17
fix(search): resolve IME inset scroll bug and keyboard flicker on back gesture

### LINKED ISSUE #4130: Search bug
### Checklist - Required

1. Klik search logo
2. Type title
3. See error

### Expected behavior

A list of songs should appear based on the title you searched for.

### Actual behavior

But what happened was that a list of previously searched songs appeared, and when I clicked the back button or the back gesture to go to the main menu, my keyboard keeps popping up and closing. So I have to press the back button on the top left. And again when I type again I just found the song at the bottom of my search history and I couldn't scroll down at all. 

The bug it's not funny

https://github.com/user-attachments/assets/41d4150d-c535-44f9-92d6-125d762daaf1

### Logs

I can't use shizuku 

### Metrolist version

13.6.0

### Android version

Android 8

### Additional information

_No response_

SANDRO00O:
## Problem

Three related bugs in the online search screen:

1. **Cannot scroll down to results** — After typing a query, items returned
   by the YouTube API appear below search history entries but are unreachable.
   The list's `contentPadding` only accounted for the system navigation bar,
   not the IME (keyboard) height. When the keyboard is open, the bottom portion
   of the list has no extra padding, so the scrollable range ends behind the
   keyboard.

2. **Keyboard dismisses unexpectedly while results load** — The scroll listener
   was keyed on `firstVisibleItemScrollOffset`. This value changes whenever
   `animateItem()` shifts item positions during list population — before the
   user touches the screen at all. This caused `keyboardController?.hide()` to
   fire the moment results loaded, not when the user actually scrolled.

3. **Keyboard flickers on system back gesture** — The `DisposableEffect`
   lifecycle observer called `keyboardController?.hide()` on `ON_PAUSE`.
   Android already dismisses the IME during back navigation; adding an explicit
   hide races with the transition animation and produces a visible show/hide
   flicker. Users had to use the toolbar back arrow to avoid this.

## Fix

**`OnlineSearchScreen.kt`**

- Replace `firstVisibleItemScrollOffset` snapshot with `isScrollInProgress`.
  Keyboard is now hidden only when the user is actively scrolling, not during
  item load animations.

- Add `WindowInsets.ime` to the `LazyColumn` `contentPadding` via `.union()`.
  The list now reserves space equal to the keyboard height at the bottom, making
  all items reachable while the keyboard is open.

**`SearchScreen.kt`**

- Remove `keyboardController?.hide()` from the `ON_PAUSE` lifecycle branch.
  Focus is still cleared on pause; IME dismissal is left to the system, which
  handles it correctly during navigation transitions.

## Testing

- [x] Open search → type a query → results appear and list is fully scrollable
      above the keyboard
- [x] Results load without keyboard dismissing on its own
- [x] System back gesture exits search without keyboard flicker
- [x] Toolbar back arrow still works as before
- [x] Player expanded state still suppresses keyboard correctly

## Related Issues

- Closes #4130 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyboard behavior while scrolling search results.
  - Search results now account for both system bars and the on-screen keyboard, preventing content from being obscured.
  - Preserved automatic keyboard dismissal during back navigation while maintaining focus behavior during scroll-to-top transitions.


app/src/main/kotlin/com/metrolist/music/ui/screens/search/SearchScreen.kt
```diff
@@ -328,10 +328,11 @@ fun SearchScreen(
                     }
 
                     Lifecycle.Event.ON_PAUSE -> {
+                        // Do not manually hide keyboard here — Android dismisses the IME
+                        // automatically during back navigation. Hiding it explicitly races
```
kairosci: Can you trim it into a single line to make it look neater?

SANDRO00O: done.

SANDRO00O:
@nyxiereal why?

nyxiereal:
missclicked, sorry

nyxiereal:
@ZicoVanCo does this solve your issue?

ZicoVanCo:
It's still the same. When I use YouTube Music, I can scroll through the music search results. But here, it doesn't work at all.

nyxiereal:
@SANDRO00O fix pls

SANDRO00O:
@ZicoVanCo can you provide me with more information? like video or somthing

nyxiereal:
The reporter says the final patch still does not fix the affected device. `isScrollInProgress` also includes programmatic movement, so it is not a reliable signal for user drag dismissal. Please reproduce the Android 8 case, react specifically to drag interactions, verify that IME padding is applied at the scrolling container that needs it, and provide before/after device evidence or a focused UI test.

SANDRO00O:
closed. if anyone else can fix this please go on.

PR #4172, opened 2026-07-27
fix(ui): prevent NPE when home screen songs are deleted from the database

GameOn223:
## Problem

`HomeScreen` re-reads every song in the **Quick picks** and **Forgotten favorites** shelves from the database so the rows stay in sync with library/like changes:

```kotlin
val song by database
    .song(originalSong.id)
    .collectAsStateWithLifecycle(initialValue = originalSong)
```

`DatabaseDao.song()` returns `Flow<Song?>`, but the value was force-unwrapped with `!!` at 16 call sites across these two shelves. `initialValue = originalSong` only covers the first frame — as soon as the flow emits for a song that is no longer in the `song` table, every `song!!` throws `NullPointerException` and takes down the home screen.

Both shelves render from a snapshot list held in `HomeViewModel` (the `quickPicks` / `forgottenFavorites` `MutableStateFlow`s), so the rows stay composed after the underlying rows are deleted.

## How to reproduce

Log out — or use *Clear library data* — while the home screen is populated. `SyncUtils.clearAllLibraryData()` ends with:

```kotlin
safeRawQuery("DELETE FROM song WHERE dateDownload IS NULL")
```

Every non-downloaded quick pick / forgotten favorite then emits `null`, and the screen crashes.

The same crash happens on a smaller scale through `SongMenu` → *Delete uploaded song*, which calls `delete(song.song)` on a row that may still be on screen.

## Fix

Fall back to the `originalSong` that is already in hand — the same pattern `SongMenu` already uses:

```kotlin
val songState = database.song(originalSong.id).collectAsStateWithLifecycle(initialValue = originalSong)
val song = songState.value ?: originalSong
```

Applied to both shelves:

```kotlin
val updatedSong by database
    .song(originalSong.id)
    .collectAsStateWithLifecycle(initialValue = originalSong)
val song = updatedSong ?: originalSong
```

This removes all 16 `!!` in the two shelves. Behaviour is unchanged while the song exists; when it disappears the row keeps rendering the last known metadata instead of crashing.

## Testing

- `./gradlew :app:assembleFossDebug` passes.
- Confirmed the premise against an in-memory copy of the app database: with a collector active on `dao.song(id)`, deleting the row makes the flow emit `null` — the exact emission that used to reach `!!`.
- `song!!` no longer occurs anywhere in `app/src/main/kotlin`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved song data handling in Quick Picks and Forgotten Favorites.
  * Play, pause, queue, and song menu actions now continue to work reliably when database data is unavailable.
  * Prevented potential crashes caused by missing song records.


kairosci (approved):
LGTM

PR #4180, opened 2026-07-28
Android auto support

rakshit1610:
## Problem
 Voice and Android Auto search behaved inconsistently, and the Auto experience was bare compared to the phone app:

  - Saying "play X" by voice would often play the wrong song, and the *same query played different songs run-to-run* (flaky).
  - After a voice "play X", playback *stopped once the search-results list ran out* instead of continuing like YouTube Music.
  - The Android Auto home offered only flat library folders — no personalized, directly-playable suggestions like the phone home screen.

## Cause
  - **Wrong match:** voice search used YouTube.search(FILTER_SONG), which ranks by spelling similarity rather than relevance, so near-spelled titles beat the intended song.
  - **Flakiness:** the code did a fire-and-forget DB insert and then immediately read the row back; the read raced the async write, sometimes returned null, and dropped the top match — so the played song varied between runs. When the returned result was null, it would not play anything and show "SOMETHING WENT WRONG" error.
  - **Playback stopping:** voice play seeded the queue with a finite search-results list and no radio continuation, so it ended when the list did.
  - **Bare Auto home:** the browse tree exposed only library folders; Android Auto root children become the tab bar, so playable tiles can't live at the root and had never been wired into a tab.

## Solution
  - **Relevance ranking:** switch voice/Auto search to YouTube.searchSummary() (relevance-ranked), drop the "Top result" card and video uploads, and place the best online match first for voice play.
  - **Race fix:** build the MediaItem directly from the SongItem (carrying its metadata tag) instead of reading it back from the DB; the insert stays a background write. Deterministic, no dropped match.
  - **Endless radio:** after a voice "play X", seed a YouTube-Music-style radio from the top match so playback continues indefinitely.
  - **Android Auto Home tab:** add a browsable "Home" tab whose children are directly-playable song tiles for Speed dial, Quick picks, and Listen again, grouped under section headers. Tapping a tile plays that row from the tapped song.
  
## Issues I encountered while testing 
- If the voice playback doesn't start, try playing a song for the first time manually from phone or the auto screen after you install the app. This will establish the media connection between the app and gemini/auto if it's cold for some reason. Most probably, you will only need to do this once. Next time you connect your phone, the session would be remembered.
  
- If you have a native language that's not English, let's say Hindi. If you ask Gemini to play a song in Hindi like "Mashooqa bajao", it might try reaching out for the original YT music because the Gemini doesn't have a strong intent as it doesn't understand the language and translates "bajao" as "play" so it hands the query over to the YT music. In that case, you will have to say something like "Metrolist pe Mashooqa bajao" (Play Mashooqa on Metrolist) so that it catches which app to route to. This can be solved if you add Hindi as a language in Gemini app - preferably in the spoken languages section and then you don't need to explicitly mention Metrolist.

## Related Issues

- Closes #
- Related to #

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -162,9 +162,12 @@ android {
         }
         create("release") {
             storeFile = file("keystore/release.keystore")
-            storePassword = System.getenv("STORE_PASSWORD")
-            keyAlias = System.getenv("KEY_ALIAS")
-            keyPassword = System.getenv("KEY_PASSWORD")
+            // Fall back to the personal-use defaults baked into the keystore we generated,
+            // so a local release build works without exporting env vars. CI/official builds
+            // can still override via STORE_PASSWORD/KEY_ALIAS/KEY_PASSWORD.
+            storePassword = System.getenv("STORE_PASSWORD")?.takeIf { it.isNotBlank() } ?: "metrolist-release"
+            keyAlias = System.getenv("KEY_ALIAS")?.takeIf { it.isNotBlank() } ?: "metrolist"
+            keyPassword = System.getenv("KEY_PASSWORD")?.takeIf { it.isNotBlank() } ?: "metrolist-release"
         }
         getByName("debug") {
             keyAlias = "androiddebugkey"
@@ -184,6 +187,11 @@ android {
                 getDefaultProguardFile("proguard-android-optimize.txt"),
                 "proguard-rules.pro",
             )
+            // Sign with our own release key when the keystore is present (personal builds).
+            // Without it, the release APK would be unsigned and won't install.
+            if (file("keystore/release.keystore").exists()) {
+                signingConfig = signingConfigs.getByName("release")
+            }
         }
         debug {
             if (applicationIdOverride == null) {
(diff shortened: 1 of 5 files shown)


FireLion137:
First of all, I'm glad someone else besides me is interested in Android Auto hahaha
Now about the PR: 
In one month or two the devs will release the KMP version, your pr adds HOME that is basically a feature, which are limited because they are moving to the kmp version, so unfortunately this pr should be invalid for now (when kmp releases maybe adapt your code on that version).
For the voice query part which is a "fix", i actually did a pr myself some weeks ago (#4127) which does not use your approch that changes the existing code and use Youtube.searchSummary(): i just added checks for local playlist (to play playlist by its name) and other checks based on already present code that uses Youtube.search().
I tried your approch on the Desktop Head Unit (literally used your MediaLibrarySessionCallback without changes) and it works the same (tried a few songs and the result was the same on my approch and yours), so while you justified it, i could not make different results (maybe you can drop a few use cases so ill try?) and then (to me) there is no reason to change the already present code.
By the way, i saw that you build a radio queue out of your query. In my case i decided to go for playing just the song that i wanted to play (just because i like it that way). I could just add a new preference key that if it is enabled it simply does a Youtube.radio on that song but if i remember correclty, if you have the auto load more song setting enabled it should work too.
Sorry for the wall of text 😇

Issue #4175, opened 2026-07-27
UI BUG


Ronin-music:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

1.Open Metrolist
2.Choose any music
3.Click on Sleep timer button

### Expected behavior

..

### Actual behavior

The text is incorrectly positioned in Russian.

### Screenshots/Screen recordings

<img width="720" height="1451" alt="Image" src="https://github.com/user-attachments/assets/3529dacd-2a40-4d15-921e-d1669445c69e" />

### Logs

.

### Metrolist version

13.6.1

### Android version

Android 15

### Additional information

.

NoAccount1:
This bug is also present in French with Metrolist 13.6.1

PR #4186, opened 2026-07-28
fix: Fix sleep timer dialog layout

kairosci:
## Problem
The sleep timer dialog displays the 'Set as default' and 'End of song' options side by side (horizontally).

## Cause
The options in the sleep timer dialog were contained within a Row layout.

## Solution
Changed the Row layout to a Column layout in both Player and Queue sleep timer dialogs so that the options are displayed one below the other vertically.

## Testing
Verified that the sleep timer dialog displays both options correctly aligned one below the other.

## Related Issues
Related to #4175

## Summary by CodeRabbit

* **Style**
  * Updated sleep timer dialog controls to display vertically with improved spacing and alignment.
  * Applied the same layout refinement across the player and queue interfaces.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/player/Player.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/player/Player.kt
@@ -678,9 +678,9 @@ fun BottomSheetPlayer(
                         steps = (120 - 5) / 5 - 1,
                     )
 
-                    Row(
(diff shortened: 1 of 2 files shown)


kairosci:
On the emulator, it worked at 90% scale; on the phone at 125% scale, it has the same visual issue, so it's better to switch to rows instead of keeping it in columns.

kairosci:
It works the same way even at 125%

PR #4187, opened 2026-07-29
feat(player): skip sponsor and non-music segments with SponsorBlock

### LINKED ISSUE #601: Sponsorblock : Skip option sponsorship/ advertisement/ intro part of songss
### Checklist

### Feature description

Sponsorblock: Please add an option to skip intro part of songs  / dialogues in between the songs/ outros 

Youtube has different versions of same song so there isn't any way to differentiate so sponsorblock would be helpful

### Additional information

_No response_

GameOn223:
## Problem

Sponsor reads, intros, outros and other non-music stretches inside music videos play in
full, because Metrolist has no way to know they are there. #601 is the most-upvoted open
request on the tracker (25 reactions) asking for those to be skipped automatically.

> A note on the freeze: I understand the template asks for no new features right now. I
> would rather offer this than sit on it — if the freeze still applies, feel free to park
> or close it and I will rebase whenever it lifts. Equally happy to adjust the scope or
> the defaults if you want it shaped differently.

## Cause

Not a defect. `MusicService` plays each media item from start to finish and there is no
client for a segment database, so there is nothing to skip against.

## Solution

- New `:sponsorblock` module, laid out like the existing `:lrclib` and `:lastfm` modules.
    - Queries the **hash-prefix** endpoint (`/api/skipSegments/<4-hex-prefix>`), so only the
      first four characters of the SHA-256 of the video ID ever leave the device. The server
      is never told which track is playing; the response covers every video sharing that
      prefix and is filtered locally.
    - Only `actionType = skip` entries are used. Unknown categories, malformed ranges and
      community-downvoted segments (`votes < 0`) are discarded rather than trusted.
    - Overlapping and touching segments are merged and anything under 1s is dropped, so an
      intro running straight into a sponsor read becomes one clean seek instead of two
      audible ones.
- `MusicService.setupSponsorBlock()` fetches segments on media item transition and seeks
  past them during playback. It sleeps until the next segment is due instead of polling the
  player, so a track with no segments costs nothing, and the previous watcher is cancelled
  on every transition so a stale track's segments can never apply to the new one.
- Settings screen at `settings/player/sponsorblock`, reachable from Player and audio, with a
  master toggle, a skip-notification toggle and per-category switches. Toggling anything
  re-reads segments for the track already playing rather than waiting for the next one.
- **Off by default.** The network lookup only happens once a user opts in, so nothing
  changes for anyone who does not go looking for it.
- Enabled categories default to Non-music, Sponsor and Unpaid/self promotion; the rest are
  opt-in. That follows the discussion in #601, where Non-music is the category that actually
  matters for a music app.

## Testing

- `./gradlew :app:assembleFossDebug` → BUILD SUCCESSFUL, no Kotlin errors.
- `./gradlew :sponsorblock:testDebugUnitTest` → 9 tests, all passing. They cover the hash
  prefix only exposing four characters, merging of touching and overlapping segments, the
  minimum-duration filter, segment ends being exclusive (so a skip cannot immediately
  re-trigger itself), and rejection of unusable API entries.
- No database schema changes, no version bump, no markdown changes, and new strings are
  English-only in `metrolist_strings.xml`.

## Related Issues
- Closes #601

## Summary by CodeRabbit

* **New Features**
  * Added SponsorBlock integration to automatically skip configured segments during playback.
  * Added a dedicated SponsorBlock settings screen under player settings, including enable/disable and “notify on skip”.
  * Added per-category toggles (sponsor, self-promo, interaction, intro/outro, preview, filler, non-music) persisted across sessions.
  * Added localized UI text and attribution for SponsorBlock.
  * Preferences take effect immediately for the currently playing track.


GameOn223:
Thanks for the review — all three were valid, and all three are fixed in bf970c5.

**Stale wait after a seek or speed change.** The watcher was sleeping on a delay derived from the position and rate at the time it was computed, so scrubbing backward into a segment would not skip until that old delay expired. Seeks and speed changes now restart the watcher via `onPositionDiscontinuity` and `onPlaybackParametersChanged`. It reuses the segments already fetched, so there is no extra network request per skip. The watcher's own seek is flagged and ignored — otherwise it would cancel itself mid-skip and could bounce straight back into the segment it had just left.

**Notification preference not observed.** It is now part of the preference tuple, so toggling it applies to the track already playing instead of only the next one.

**CancellationException.** `getSegments` no longer routes it into a failed `Result`; it is rethrown. Swallowing it meant a cancelled lookup carried on as an ordinary "no segments" answer.

Re-verified after the changes: `:app:assembleFossDebug` BUILD SUCCESSFUL with no Kotlin errors, and `:sponsorblock:testDebugUnitTest` at 10/10 — I added a case covering a backward seek into an already-passed segment.

nyxiereal:
image

PR #4214, opened 2026-08-04
Changed the logic of the queue created by AA search

FireLion137:
Now it checks for AutoRadioQueueKey instead of creating a nonsensical queue of search results.

## Problem
The Android Auto search page, when clicking on a song, creates a queue of search Results. This is in my opinion nonsensical, why, for example, should i want a queue of all the variants of Bohemian Rhapsody if i search precisely that title and click on the song i want? I should expect it to start that song and a queue of related song (aka radio songs).

## Cause
Simply as i said in the problem, the code was starting a searchResults.map

## Solution
Changed the logic of the queue created by AA search
- Now it checks for the already existing AutoRadioQueue setting to create a radio queue, if that is disabled, it instead plays only that song and not the search results map as before.
- Also, by changing the currentQueue of MusicService with a little new method, now the queue respect the other Load More Songs setting.

## Testing
Tested by using the AA Desktop Head Unit on Android Studio.

## Summary by CodeRabbit

* **New Features**
  * Search-result playback now selects the requested song before building the queue.
  * When enabled, automatic radio can populate the playback queue and begin at the selected song.
  * Playback falls back to the selected song when radio content is unavailable.
  * Queue handling now better distinguishes original songs from automatically added tracks.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/playback/MediaLibrarySessionCallback.kt
+++ b/app/src/main/kotlin/com/metrolist/music/playback/MediaLibrarySessionCallback.kt
@@ -63,9 +63,13 @@ import kotlinx.coroutines.plus
 import javax.inject.Inject
 import com.metrolist.music.constants.AndroidAutoSectionsOrderKey
 import com.metrolist.music.constants.AndroidAutoYouTubePlaylistsKey
+import com.metrolist.music.constants.AutoRadioQueueKey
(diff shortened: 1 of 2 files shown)


FireLion137:
@nyxiereal can you please review when you can? thanks

nyxiereal:
lgtm

Issue #3703, opened 2026-05-05
Metrolist crashes Android Auto


miguelcanosantana:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

The entire Android Auto UI crashes, making you having to connect and disconnect from the car.

Since last versions, maybe related to this issue: https://github.com/MetrolistGroup/Metrolist/issues/3550
When a song fails to load and crashes metrolist using it in android auto. The entire android auto fails and you have to connect and disconnect the phone manually from the car, which you cannot do driving for safety.

Steps:
- Play multiple songs in android auto until one stutters or refuses to load
- Aftter the stuttering or completely refusing to load, wait for a few seconds. Android auto ui will become unresponsive.
- Android auto crashes, al you have to connect the phone again to restart it

### Expected behavior

When there's an error in a song, the entire Android Auto stack shouldn't crash. Only metrolist. 

### Actual behavior

The whole android auto system crashes, making you having to unlock the phone to restart it and disconnect and connect from the car. 

### Logs

Its not safe to log whilst driving. 

### Metrolist version

13.4.2

### Android version

Android 16

miguelcanosantana:
Just to clarify, it's not duplicated of #1756 above.

Most of the songs play normally, but it happens randomly. Also, there's no error message. Android auto disconnects, which is the difference.

idont85:
My android auto crashes as well (often) while listening to Metrolist, even if the app is just open. I haven't narrowed it down to just Metrolist, but your issue seems similar. Are you on OneUI perchance?

miguelcanosantana:
Yes, I'm using OneUI, might that be related in any way?

fyreswan:
I am having this behaviour occur on a google pixel phone (not rooted) with the latest android update as well

FabianLars:
[...]

Issue #3478, opened 2026-04-07
Android auto


ahmadw13:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

App doesn't appear on android auto, in customise launcher settings, it also doesn't appear there so I add it 

### Expected behavior

It should appear

### Actual behavior

Doesn't appear in android Auto 

### Logs

No logs 

### Metrolist version

13.4.0

### Android version

16

nyxiereal:
You didn't install it properly

Issue #1806, opened 2025-10-12
Android Auto


Vimlesh-024:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

Hey i have used the app through Android auto in my car, the app logo doesn't show's up there on the screen, it's just a black circle and nothing in it.

### Expected behavior

App logo should be visible 

### Actual behavior

App logo is not visible 

### Logs

Don't have it

### Metrolist version

12.7.0

### Android version

Android 12 maybe

jimbo601:
Make sure you have "Unknown Sources" on in the Android Auto developer sources. 
[Try here](https://developer.android.com/training/cars/testing#unknown-sources) for more info on this option, but you'll have to navigate into the Android Auto settings from your app page in your settings app, then tap the version like 6 times to enable developer options. Under the meatball menu for more options, there should be developer settings once it's turned on, somewhere down that list there should be an option for "Unknown Sources," once that's checked you should be able to add it from the Android Auto home screen customize settings.

PR #4255, opened 2026-08-14
Android auto

chuppito:
Fixes Android Auto not appearing / hanging in the app launcher
(#1806, #3478, #3703)

onCreate() called startForeground() unconditionally and returned
early (skipping MediaLibrarySession creation) if the promotion was
denied. This also runs when the system only binds the service for
browsing (Android Auto, Assistant, etc.) without ever calling
startForegroundService() — a case where foreground denial is
expected and not an error. Skipping session creation in that case
left browse-only clients with no valid session to connect to.

This moves the foreground promotion attempt to best-effort at
onCreate() and always proceeds to create the MediaLibrarySession.
The strict "must promote within timeout" requirement remains
enforced separately in onStartCommand(), which only fires for a
real startForegroundService() call triggered by actual playback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback service startup reliability when foreground promotion is unavailable.
  * Browse-only connections, including Android Auto, can continue initializing without interruption.
  * Foreground playback remains enforced when playback starts.


nyxiereal:
Continues synchronous service startup after foreground promotion fails. This can miss Android’s foreground-service deadline and break playback startup. Needs a lifecycle redesign.

chuppito:
Thanks for flagging this, that's a fair point.

You're right that if the foreground promotion in `onCreate()` fails, the rest of `onCreate()`'s setup still runs synchronously before `onStartCommand()` gets a chance to retry via `ensureForegroundWithLatestNotificationOrStop()`. If that setup work is slow enough (e.g. the `dataStore.data.first()` read on some OEMs), it could eat into the 5s `startForegroundService()` window before the retry even happens, so the failure mode you're describing is real.

In practice this specific race should be narrow: real playback-triggered starts (media button, foreground UI) are normally exempt from the foreground-start restriction, so a denial at `onCreate()` almost always means this was actually a background/browse-only bind (Android Auto, Assistant, etc.) rather than a genuine playback start - which is the case this fix targets. But "almost always" isn't "never," so I don't want to wave it off.

A proper fix would mean restructuring `onCreate()` so the heavy setup (player/session/listener init - it's a big chunk of the method) can run asynchronously on the failure path, letting `onStartCommand()`'s retry fire immediately instead of waiting on that work. That's a real lifecycle change, not a small patch, and I'd rather not push something that size without being able to properly test it against the failure path (no easy way for me to reliably repro a foreground-denial-during-real-playback scenario locally).

I'd like to treat this as a follow-up rather than blocking this PR on it, since this fix addresses a separate, confirmed bug (the app being invisible in Android Auto). Happy to open a tracking issue for the lifecycle rework if that works for you.

nyxiereal:
Thanks for confirming the race. I do not think it is safe to defer it: `onCreate()` cannot tell whether creation came from a bind or `startForegroundService()`, and `onStartCommand()` cannot retry promotion until `onCreate()` returns, after the synchronous initialization that can consume the deadline. The safe shape is a lightweight/non-blocking `onCreate()` that still allows a browse-only bind to initialize its session, with immediate foreground enforcement once a started-service path is known. I would keep this blocked until that lifecycle split is implemented and tested; a tracking issue is still useful for designing it.

chuppito:
That's a clear and correct framing, thanks. Agreed on all three points: `onCreate()` has no way to know why it was invoked, `onStartCommand()` can't retry until `onCreate()` returns, and synchronous init in between can consume the deadline either way. Deferring the blocking read only narrows the window, it doesn't close it.

I'll leave this blocked. Doing the lifecycle split properly means going through everything that currently assumes the player/session/listeners are ready by the time `onCreate()` returns, so it's not something I want to push without being able to actually observe the failure path (logcat) rather than guessing from black-box testing. I'll get that set up and come back to this once I can test it properly rather than rush a change I can't verify.

I'll open a tracking issue to sketch out the split (minimal browse-capable `onCreate()` + foreground promotion as the first thing on the started-service path) so the design discussion isn't lost. Thanks for pushing on this.

nyxiereal:
closed as stale and harmful, redesign needed

Issue #3561, opened 2026-04-14
Weird Bug when changing list order in custom order format


Elekrys:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

1.Create a playlist with some songs 
2.Put the playlkst in custom order 
3.Add a new song and drag it all the way to the top and you can see it replace it with another song from the bottom

### Expected behavior

It should replace the original song when dragged to the top instead of another song

### Actual behavior

The song gets replaced by some other irrelevant song from the bottom

### Screenshots/Screen recordings

https://github.com/user-attachments/assets/350003f3-b58d-49b1-ba29-f017182a01a8

### Logs

[12_03-23-00-06_022.log](https://github.com/user-attachments/files/26730558/12_03-23-00-06_022.log)

### Metrolist version

13.4.1

### Android version

16

### Additional information

Please fix it ASAP  Its a really  annoying bug

johannesbrauer:
I will look into it, think it's caused somewhere in the sync process. I'm going to look how/if we can fix this.

johannesbrauer:
Yeah it's caused somewhere in the sync / especially auto sync process.

johannesbrauer:
I'm going to fix this, but I need a bit time.

johannesbrauer:
@Elekrys  doesn't happen to me anymore, could you recheck?

Koppany69:
It still happens to me

johannesbrauer:
> It still happens to me

Proof with current version please.

Koppany69:
> > It still happens to me
> 
> Proof with current version please.

The playlist's first part is fine, but the middle and last parts are buggy (version 13.5.0): https://github.com/user-attachments/assets/a765b705-a24a-4db3-ad00-e6ebc9895dfe

johannesbrauer:
> > > It still happens to me
> > 
> > Proof with current version please.
> 
[...]

PR #4267, opened 2026-08-17
fix(sync): keep local playlist order when syncing with YouTube

fatuhsa:
## Problem
Reordering songs in a synced playlist never sticks: after moving a song (e.g. from position 5 to position 1), the playlist snaps back to the original order — even with the lock button unlocked.

## Cause
`executeSyncPlaylist` compared the remote and local song ID lists by **order** (`remoteIds == localIds`). A reorder is pushed to YouTube Music from the playlist screen, but when that push hasn't propagated yet (or fails), the remote order still differs from local. The sync then treated the playlist as out of sync, wiped it (`clearPlaylist`) and rebuilt it in the remote order — reverting the user's reorder. Same root cause as #3561 / #1318 / #1231 (sync-related reorder reverts).

## Solution
- Compare song ID **sets** instead of ordered lists: when both sides contain the same songs, the local order is authoritative and the sync does nothing.
- When membership genuinely differs, **merge** instead of wipe-and-rebuild:
  - songs present on both sides keep their local position
  - songs removed remotely (and not downloaded) are dropped
  - new remote songs are appended at the end in remote relative order
  - downloaded songs are preserved
  - positions are re-normalized so future inserts can't collide

## Testing
- No local build per repo workflow; CI (`build_pr.yml` / `build_quick.yml`) verifies compilation.
- Manual test to confirm: unlock a synced playlist, move a song to a different position, then refresh / restart the app — the order should stick.

## Related Issues
- Related to #3561
- Related to #1318
- Related to #1231

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playlist synchronization to preserve local song ordering when membership is unchanged.
  * Retained downloaded songs while removing songs no longer present remotely.
  * Added new remote songs without unnecessarily rebuilding the playlist.
  * Ensured playlist positions remain consistent after synchronization.
  * Prevented unnecessary metadata updates when playlist membership has not changed.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
@@ -1532,51 +1532,75 @@ class SyncUtils @Inject constructor(
                     val remoteIds = songs.map { it.id }
                     val localIds = database.playlistSongIds(playlistId)
 
-                    if (remoteIds == localIds) {
-                        Timber.d("syncPlaylist: Local and remote are in sync, no changes needed")
+                    // Compare as sets rather than ordered lists: reorders are pushed to
+                    // YouTube Music from the playlist screen, so when both sides contain the
+                    // same songs the local order is authoritative and must not be overwritten
+                    // by the remote order (which may not have caught up with the reorder yet).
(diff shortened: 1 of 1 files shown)


fatuhsa:
Closing for now — testing the build on a fork before finalizing.

PR #4268, opened 2026-08-17
fix(sync): keep local playlist order when syncing with YouTube

fatuhsa:
Replaces the closed #4267 (its head branch was squashed into a single commit and force-pushed, so GitHub blocks reopening it). Same change, clean history.

## Problem
Reordering songs in a synced playlist never sticks: after moving a song (e.g. from position 5 to position 1), the playlist snaps back to the original order — even with the lock button unlocked.

## Cause
`executeSyncPlaylist` compared the remote and local song ID lists by **order** (`remoteIds == localIds`). A reorder is pushed to YouTube Music from the playlist screen, but when that push hasn't propagated yet (or fails), the remote order still differs from local. The sync then treated the playlist as out of sync, wiped it (`clearPlaylist`) and rebuilt it in the remote order — reverting the user's reorder. Same root cause as #3561 / #1318 / #1231 (sync-related reorder reverts).

## Solution
- Compare song ID **sets** instead of ordered lists: when both sides contain the same songs, the local order is authoritative and the sync does nothing.
- When membership genuinely differs, **merge** instead of wipe-and-rebuild:
  - songs present on both sides keep their local position
  - songs removed remotely (and not downloaded) are dropped
  - new remote songs are appended at the end in remote relative order
  - downloaded songs are preserved
  - positions are re-normalized so future inserts can't collide

## Testing
- Build verified in CI on the fork: `:app:assembleFossDebug` passes (run [31986931101](https://github.com/fatuhsa/Metrolist/actions/runs/31986931101)).
- Manual test to confirm: unlock a synced playlist, move a song to a different position, then refresh / restart the app — the order should stick.

## Related Issues
- Related to #3561
- Related to #1318
- Related to #1231

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty remote playlists now synchronize correctly.
  * Playlist synchronization preserves local song ordering when memberships match.
  * Downloaded songs are retained while applying remote membership changes.
  * Newly available remote songs are appended and playlist positions are compacted.
  * Missing artist information is repaired during song and playlist synchronization.
  * Artist subscriptions now resolve aliased artist names correctly.
  * Library cleanup preserves downloaded songs and removes orphaned artist entries.
  * Playlist timestamps update only when membership changes.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt
@@ -1560,56 +1560,87 @@ class SyncUtils @Inject constructor(
                     val localIds = database.playlistSongIds(playlistId)
                     val songIdsWithoutArtists = database.playlistSongIdsWithoutArtists(playlistId).toSet()
 
-                    if (remoteIds == localIds) {
+                    // Compare as sets rather than ordered lists: reorders are pushed to
+                    // YouTube Music from the playlist screen, so when both sides contain the
+                    // same songs the local order is authoritative and must not be overwritten
+                    // by the remote order (which may not have caught up with the reorder yet).
+                    if (remoteIds.toSet() == localIds.toSet()) {
+                        // Even when membership matches, repair any songs missing artist metadata.
                         val metadataRepairs = songs.filter {
                             it.id in songIdsWithoutArtists && it.artists.isNotEmpty()
                         }
                         if (metadataRepairs.isNotEmpty()) {
                             database.withTransaction {
-                                metadataRepairs.forEach(::insert)
+                                metadataRepairs.forEach(database::insert)
                             }
                         }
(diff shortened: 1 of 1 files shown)


nyxiereal:
A valid empty remote playlist still returns before reconciliation, so clearing a playlist on YouTube leaves local non-downloaded songs behind. Please represent fetch failure separately from an empty successful result, reconcile the empty case, and add coverage for empty playlists, downloaded-song preservation, duplicate IDs, and final position normalization.

fatuhsa:
@coderabbitai review

The fork_ci.yml file that triggered the previous CHANGES_REQUESTED has been removed from this branch. The PR now only contains changes to `app/src/main/kotlin/com/metrolist/music/utils/SyncUtils.kt`. Please re-review.

nyxiereal:
This ignores legitimate remote reorders, and also conflicts with safe reconciliation.

PR #4269, opened 2026-08-17
fix(lyrics): stop background vocals cutting a line's last word short

### LINKED ISSUE #4204: last word of the verse randomly plays while using {bg} on custom made lyrics
### Checklist - Required

I made a tool to make WBW lyrics. I updated it to also support background voices. When I import them on Metrolist or any other client, the last word of the verse in which there is a background voice plays before it has to

### Expected behavior

last word should play when it's its time

### Actual behavior

it plays before it has to

https://youtube.com/shorts/5nz6CXxnzms (couldn't upload it)

### Logs

no log 

### Metrolist version

13.6.1

### Android version

Android 17 QPR2 Beta 1

### Additional information

if you want to try the tool it's in my repo: https://github.com/0xygen33/lyricsyncher2
completely open source. byebye

GameOn223:
## Problem

In word-by-word lyrics, the last word of a line that has background vocals finishes early —
it renders as already sung before its turn. Lines without background vocals are unaffected,
which is what makes it look random.

## Cause

`LyricsUtils.getNextLineStartTime()` derives the last word's end time from the start of the
next line, and it explicitly accepted a background line as that "next line".

Background vocals overlap the line they accompany rather than following it, so a background
line does not mark where the previous line ends. Taking its start time cuts the last word
short, and when the background begins *before* that word does, the word ends up with an end
time earlier than its own start — so it is drawn as already finished.

```
[00:10.00]<00:10.00>Main <00:10.50>line
[bg: <00:10.20>ooh<00:11.00>]
[00:12.00]<00:12.00>Next <00:12.40>line
```

Here `line` starts at 10.50 but was given an end time of 10.20.

## Solution

- `getNextLineStartTime()` skips background lines when looking for the next line, so the last
  word now runs to the next *main* line instead of being cut off by an overlapping background.
- Added `isBackgroundLine()` covering both spellings the parser already accepts: a whole
  `[bg: ...]` line, and a timestamped line whose content begins with `{bg}`.

Behaviour is otherwise unchanged: if the first non-background line is not a rich-sync line the
function still returns `null` and the existing fallback applies.

## Testing

Added `LyricsBackgroundTimingTest`. I confirmed the tests fail before the change and pass after.

Before:

```
last word of a line with background vocals does not end before it starts
    java.lang.AssertionError: last word ends at 10.2 but starts at 10.5
last word runs to the next main line rather than to the background line
    java.lang.AssertionError: expected:<12.0> but was:<10.2>
4 tests completed, 2 failed
```

After: 5/5 pass. The two control cases — a line followed by an ordinary line, and the
background line still parsing with its own timings — pass both before and after, so the change
is scoped to the overlapping-background case.

- `./gradlew :app:assembleFossDebug` — BUILD SUCCESSFUL.
- `./gradlew :app:testFossDebugUnitTest` — 60 tests, 4 failing. Those 4 are all in
  `YouTubeUtilsTest` (thumbnail URL rewriting) and are **pre-existing on `main`**: I stashed this
  change and re-ran that class on a clean tree, which reproduced the same 4 failures, so they are
  unrelated to this PR. Happy to open a separate issue for them if that would help.

## Related Issues
- Closes #4204

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lyric word timing when background vocals overlap the main lyrics.
  * Prevented background-vocal timestamps and continuation lines from prematurely ending preceding words.
  * Added support for both recognized background-vocal lyric formats.
  * Preserved correct timing for standard consecutive lyric lines.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/lyrics/LyricsUtils.kt
+++ b/app/src/main/kotlin/com/metrolist/music/lyrics/LyricsUtils.kt
@@ -664,34 +664,35 @@ object LyricsUtils {
      * Get the start time of the next line for calculating the last word's end time
      */
     private fun getNextLineStartTime(currentIndex: Int, allLines: List<String>): Double? {
-        if (currentIndex + 1 >= allLines.size) return null
-
-        val nextLine = allLines[currentIndex + 1].trim()
-        
-        // Try standard rich sync line
-        val matchResult = RICH_SYNC_LINE_REGEX.matchEntire(nextLine)
-        if (matchResult != null) {
-            val minutes = matchResult.groupValues[1].toLongOrNull() ?: return null
-            val seconds = matchResult.groupValues[2].toLongOrNull() ?: return null
-            val fraction = matchResult.groupValues[3].toLongOrNull() ?: 0L
-
-            val fractionPart = if (matchResult.groupValues[3].length == 3) fraction / 1000.0 else fraction / 100.0
-            return minutes * 60.0 + seconds + fractionPart
+        // Background vocals overlap the line they accompany rather than following it,
+        // so a background line does not mark where the previous line ends. Using one
+        // cut the last word short, and when the background started before that word
+        // did, the word was handed an end time earlier than its own start.
+        var nextIndex = currentIndex + 1
+        while (nextIndex < allLines.size && isBackgroundLine(allLines[nextIndex].trim())) {
+            nextIndex++
         }
-        
-        // Try background line
-        val bgMatch = PAXSENIX_BG_LINE_REGEX.matchEntire(nextLine)
-        if (bgMatch != null) {
-            val content = bgMatch.groupValues[1]
-            val wordMatch = RICH_SYNC_WORD_REGEX.find(content) ?: return null
-            val minutes = wordMatch.groupValues[1].toLongOrNull() ?: return null
-            val seconds = wordMatch.groupValues[2].toLongOrNull() ?: return null
-            val fraction = wordMatch.groupValues[3].toLongOrNull() ?: 0L
-            val fractionPart = if (wordMatch.groupValues[3].length == 3) fraction / 1000.0 else fraction / 100.0
-            return minutes * 60.0 + seconds + fractionPart
-        }
+        if (nextIndex >= allLines.size) return null
 
-        return null
+        val nextLine = allLines[nextIndex].trim()
+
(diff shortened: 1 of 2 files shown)


GameOn223:
Good catch — that one was valid, fixed in c5af77f.

`parseRichSyncLyrics` falls back to a standalone `<word:start:end>` line for a `[bg: ...]` line that has no inline timestamps, so the scan stepped over the marker line but stopped on the continuation. `RICH_SYNC_LINE_REGEX` then failed to match it and the preceding line's last word fell back to `start + 0.5s` instead of running to the next main line. The continuation is now stepped over together with the background line it belongs to.

Confirmed the same way as the original fix — the new case fails before the change and passes after:

```
before: a background line whose timings sit on a continuation line is skipped whole
        java.lang.AssertionError: expected:<12.0> but was:<11.0>   (11.0 = 10.5 + 0.5 fallback)
after:  6/6 pass
```

`./gradlew :app:assembleFossDebug` — BUILD SUCCESSFUL.

nyxiereal:
The background-line timing fix matches the parser behavior and the added regression test covers the important continuation case. I found no functional blocker; the remaining validation gap was the pending Android build, which I have now approved. Please confirm it passes before merge.

GameOn223:
Thanks for approving the run — confirming it's green.

`Build PR` passed on the PR head `c5af77f`: [run 32009058265](https://github.com/MetrolistGroup/Metrolist/actions/runs/32009058265) — `build` job success in 2m36s. CodeRabbit's latest pass also came back with no actionable comments.

Locally on the same commit: `:app:assembleFossDebug` BUILD SUCCESSFUL, and `:app:testFossDebugUnitTest --tests "com.metrolist.music.lyrics.*"` is 6/6.

One thing worth flagging since it's unrelated to this PR: the full `:app:testFossDebugUnitTest` run has 4 failures in `YouTubeUtilsTest` (thumbnail URL rewriting). I stashed this branch and re-ran that class on a clean `main` and reproduced the same 4, so they're pre-existing rather than introduced here. Happy to open a separate issue or a follow-up PR for those if useful.

nyxiereal:
Thanks for verifying. No separate issue is needed for those four failures: they are the pre-existing `YouTubeUtilsTest` mismatches addressed by #4277, whose current head now has a green build and updated regression coverage. This PR remains clean and ready from my side.

Issue #4247, opened 2026-08-10
Informe de fallos de Metrolist


makro17:
Metrolist Crash Report
==================================================

Manufacturer: Google
Device: Pixel 8a
Android version: 17 (SDK 37)
App version: 13.6.3 (152)

==================================================
Stacktrace:
==================================================

java.lang.OutOfMemoryError: Failed to allocate a 424 byte allocation with 15424 free bytes and 15KB until OOM, target footprint 268435456, growth limit 268435456; giving up on allocation because <1% of heap free after GC.
	at android.database.CursorWindow.nativeGetString(Native Method)
	at android.database.CursorWindow.getString(CursorWindow.java:505)
	at android.database.AbstractWindowedCursor.getString(AbstractWindowedCursor.java:54)
	at androidx.sqlite.driver.SupportSQLiteStatement$RowSQLiteStatement.getText(r8-map-id-bd199f5e4b39be09f62501957ec461c29a6934ae32e93a69c48b8de9ea8c764f:11)
	at com.metrolist.music.db.DatabaseDao_Impl.__fetchRelationshipformatAscomMetrolistMusicDbEntitiesFormatEntity(r8-map-id-bd199f5e4b39be09f62501957ec461c29a6934ae32e93a69c48b8de9ea8c764f:225)
	at com.metrolist.music.db.DatabaseDao_Impl$$ExternalSyntheticLambda29.invoke(r8-map-id-bd199f5e4b39be09f62501957ec461c29a6934ae32e93a69c48b8de9ea8c764f:45)
	at kotlin.math.MathKt.recursiveFetchArrayMap(dex-id-68b620342ef7b71a38b146a38a808ff572bec62d:58)
	at com.metrolist.music.db.DatabaseDao_Impl.__fetchRelationshipformatAscomMetrolistMusicDbEntitiesFormatEntity(r8-map-id-bd199f5e4b39be09f62501957ec461c29a6934ae32e93a69c48b8de9ea8c764f:35)
	at com.metrolist.music.db.DatabaseDao_Impl$$ExternalSyntheticLambda24.invoke$com$metrolist$music$db$DatabaseDao_Impl$$ExternalSyntheticLambda175(r8-map-id-bd199f5e4b39be09f62501957ec461c29a6934ae32e93a69c48b8de9ea8c764f:329)
[...]

Issue #4094, opened 2026-07-04
Crash when clicking three-dots button on a song in playlist


duggavo:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

1. Open a playlist (my playlist has 160 songs, if that matters)
2. Click the three dots on ANY song
3. It crashes

### Expected behavior

It shouldn't crash

### Actual behavior

It crashes

### Logs
```

Metrolist Crash Report
==================================================

Manufacturer: motorola
Device: moto g(8) power
Android version: 11 (SDK 30)
App version: 13.6.0 (149)

==================================================
Stacktrace:
==================================================

[...]

PR #4270, opened 2026-08-17
fix(cache): stop polling the entire media cache every second

makro17:
## Problem

Repeated `OutOfMemoryError` crashes, and on older Android a
`SQLiteException: too many SQL variables`, both traced to the Cache Playlist.

The crashes appear on screens unrelated to the Cache Playlist, often minutes after the user last
touched it, and reproduce on libraries of well under 1000 songs.

## Cause

`CachePlaylistViewModel` runs an unbounded polling loop from its `init` block:

```kotlin
while (true) {
    val candidateIds = playerCache.keys.toSet() + downloadCache.keys.toSet()
    val songs = database.getSongsByIds(candidateIds.toList())
    ...
    delay(1000)
}
```

`playerCache.keys` is every entry in the media3 `SimpleCache` — every track ever streamed, not the
user's library — and it grows without bound. Those ids go to `DatabaseDao.getSongsByIds`, a
`@Transaction` query returning `List<Song>`; `Song` declares four `@Relation` fields, so Room issues
four extra relation fetches and rebuilds a large object graph. Once per second, forever.

The allocation rate is what exhausts the heap, not the library size, which is why a small library
still crashes. The ANR is a consequence: with the heap full the process is in continuous GC and the
main thread misses the input-dispatch deadline.

The trigger is `SongMenu.kt`'s `hiltViewModel<CachePlaylistViewModel>()`. Opening the three-dots
menu of **any** song constructs the ViewModel and starts the loop, and the ViewModel outlives the
menu — which is why the crash surfaces later, elsewhere.

The same call shape fails differently depending on the SQLite bind-variable limit, 999 on older
Android and 32766 on current ones: an exception on old devices, heap exhaustion on new ones. That is
the same defect reported in #4094.

## Solution

Invert the direction of the query: instead of "every cache key, find its song", ask "every song
flagged as cached, is it still cached". `dateDownload != null` is already the authoritative marker
for Cache Playlist membership — only `MusicService.markCachedIfFullyDownloaded` sets it, once a
track has genuinely finished playing — and that set is small and bounded.

- Add a DAO query exposing the flagged set as a `Flow`. No schema change.
- Rewrite `CachePlaylistViewModel` to derive `cachedSongs` from that `Flow` combined with the
  existing `HideExplicitKey` and `HideVideoSongsKey` preference flows, via
  `stateIn(viewModelScope, SharingStarted.WhileSubscribed(5_000), emptyList())`. It recomputes on
  database invalidation instead of on a timer, and does nothing while nothing is collecting.
- Keep the self-healing behaviour: a flagged song whose bytes are gone from both caches has its
  `dateDownload` cleared. This now runs over the small flagged set instead of the whole cache.
- Delete `DatabaseDao.getSongsByIds`. The ViewModel was its only caller, and leaving an unbounded
  `IN (:songIds)` helper in the DAO invites the same defect back.

`SongMenu` keeps its `hiltViewModel` call, since it needs `removeSongFromCache`, but with
`WhileSubscribed` the ViewModel no longer does work merely by being constructed, so opening a menu
stops being a trigger.

## Testing

- `./gradlew :app:assembleFossDebug` and `./gradlew :app:testFossDebugUnitTest`.
- The staleness rule is extracted as a pure function and covered by unit tests (7 cases).
- On device: constructing the ViewModel issues no query; opening a song's three-dots menu repeatedly
  leaves heap usage flat under `adb shell dumpsys meminfo`, where before it climbed steadily.
- The Cache Playlist still lists the same songs, and still drops songs whose cached data has been
  evicted.

## Related Issues

- Related to #4247
- Related to #4094

## Summary by CodeRabbit

- **Improvements**
  - Playlist caching now updates automatically as song downloads and preferences change.
  - Cache status is refreshed more reliably, including when songs are removed or become stale.
  - Downloaded songs are retained appropriately, even when content size information is unavailable.

- **Bug Fixes**
  - Improved handling of outdated cache entries and download flags.
  - Preserved playlist order and ensured empty playlists are handled correctly.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/db/DatabaseDao.kt
+++ b/app/src/main/kotlin/com/metrolist/music/db/DatabaseDao.kt
@@ -698,8 +698,8 @@ interface DatabaseDao {
     fun getSongByIdBlocking(songId: String): Song?
 
     @Transaction
-    @Query("SELECT * FROM song WHERE id IN (:songIds)")
-    suspend fun getSongsByIds(songIds: List<String>): List<Song>
+    @Query("SELECT * FROM song WHERE dateDownload IS NOT NULL")
+    fun cachePlaylistSongs(): Flow<List<Song>>
 
 
     @Transaction
--- a/app/src/main/kotlin/com/metrolist/music/viewmodels/CachePlaylistViewModel.kt
+++ b/app/src/main/kotlin/com/metrolist/music/viewmodels/CachePlaylistViewModel.kt
@@ -97,3 +97,39 @@ class CachePlaylistViewModel
             playerCache.removeResource(songId)
         }
     }
+
+/**
+ * Result of checking which Cache Playlist entries still have their bytes on disk.
+ */
+internal data class CachedSongPartition(
+    val stillCached: List<Song>,
+    val stale: List<Song>,
+)
+
+/**
+ * Splits songs flagged as belonging to the Cache Playlist into those whose data is still
+ * present and those whose flag must be cleared.
+ *
+ * A downloaded song is always considered present: its file is managed by the download service
+ * rather than the player cache. A song whose content length is unknown cannot be checked, so it
+ * is treated as stale rather than assumed present.
+ *
+ * [isCached] receives the song id and its content length, and reports whether the complete file
+ * is available in either cache.
+ */
+internal fun partitionCachedSongs(
+    flagged: List<Song>,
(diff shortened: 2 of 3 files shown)


nyxiereal:
The bounded reactive cache query is a good replacement for polling the complete cache every second. I found no functional blocker; the remaining validation gap was the pending Android build, which I have now approved. Please confirm it passes before merge.

PR #4271, opened 2026-08-17
fix(playlist): survive a failed YouTube playlist creation

makro17:
## Problem

Creating a playlist with the "sync with YouTube" switch enabled kills the app outright when the
request fails:

```
java.net.UnknownHostException: Unable to resolve host "music.youtube.com"
```

The playlist is then not created at all, not even locally, so the user loses the work entirely.

This is not specific to being offline — any network or API failure crashes identically. Losing the
network between arming the switch and confirming the dialog reproduces it deterministically.

## Cause

`YouTube.createPlaylist` is the only playlist function in `YouTube.kt` that is neither `suspend` nor
wrapped in `runCatching` — its neighbours `renamePlaylist`, `moveSongPlaylist` and the rest all
return a `Result`:

```kotlin
fun createPlaylist(title: String) =
    runBlocking {
        innerTube.createPlaylist(WEB_REMIX, title).body<CreatePlaylistResponse>().playlistId
    }
```

`CreatePlaylistDialog` calls it inside a bare `launch` on the dialog's own
`rememberCoroutineScope()`. A bare `launch` has no handler, so the exception reaches the thread's
uncaught exception handler and terminates the process. `withRetry` inside `InnerTube.createPlaylist`
retries first and then rethrows, so the crash is merely delayed.

The local `insert` sits **after** the call that throws, which is why the playlist is lost rather
than surviving as an unsynced one.

## Solution

Write the playlist locally first and offer it to YouTube afterwards, from `SyncUtils` and its
application-lifetime scope rather than the dialog's.

- `YouTube.createPlaylist` becomes a `suspend fun` returning `Result<String>`, in line with every
  sibling in the file, and drops its `runBlocking`. `CreatePlaylistDialog` is its only caller.
- The "sync requested, not yet created" state is spelled with the existing `isAutoSync` column and
  a null `browseId`, so it needs **no schema change**. That column was already read by the auto-sync
  filter and never written by anything in the Kotlin sources, so the filter matched nothing; it now
  has its first real input. Because that filter also requires `browseId != null`, a pending playlist
  cannot be picked up half-created.
- A later playlist sync creates the pending ones on YouTube and writes the `browseId` back, reusing
  the existing `withRetry` and the playlist-edit throttle rather than adding a second retry
  mechanism. A guard ensures the dialog's own upload and a concurrent sync cannot both create it.
- One new string in `metrolist_strings.xml` tells the user the playlist exists locally and will
  reach YouTube on the next sync.

**A second, silent defect in the same dialog is fixed by the same change.** `onPlaylistCreated`
never fired: `TextFieldDialog` calls `onDismiss()` *before* `onDone(...)`, so the dialog leaves
composition, its `rememberCoroutineScope()` is cancelled, and the callback's
`withContext(Dispatchers.Main)` resumes on an already-cancelled job. Both of its consumers,
`LibraryPlaylistsScreen` and `LibraryMixScreen`, were dead code. Moving the work off the composition
scope revives them. The callback is still delivered on the main thread, because its callers
navigate.

## Testing

- `./gradlew :app:assembleFossDebug`, `./gradlew :app:testFossDebugUnitTest` and
  `./gradlew :innertube:testDebugUnitTest`.
- The pending-state predicate is a pure function covered by unit tests (5 cases).
- On device, arming the switch with the network up, cutting WiFi, then confirming:
  - the process survives, where the current release dies on the uncaught `UnknownHostException`;
  - the playlist is created locally with a null `browseId` and `isAutoSync` set, and the toast
    states it will be added to YouTube on the next sync;
  - confirming the dialog navigates into the new playlist, which is `onPlaylistCreated` firing.
- With the network restored, the next playlist sync creates it on YouTube and writes the `browseId`
  back. Three further syncs, one after a cold restart, created no duplicate: one local row, one
  remote playlist.
- A playlist created with the switch off keeps `isAutoSync` false and is never uploaded.

Not verified on device: the signed-out branch, which cannot arm the switch at all and therefore
takes the local-only path that this change makes survivable.

## Summary by CodeRabbit

- **New Features**
  - Device-only playlists can be added to YouTube during the next synchronization.
  - Playlist creation now supports retries and recovery when requests are interrupted.
  - Synchronization preserves downloaded songs and repairs missing song and artist metadata.
  - Improved detection of YouTube sign-in status.

- **Bug Fixes**
  - Prevented duplicate playlist creation and preserved playlist ordering during synchronization.
  - Improved handling of playlist cleanup and remote playlist matching.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/component/CreatePlaylistDialog.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/component/CreatePlaylistDialog.kt
@@ -32,6 +32,7 @@ import com.metrolist.music.R
 import com.metrolist.music.constants.InnerTubeCookieKey
 import com.metrolist.music.db.entities.PlaylistEntity
 import com.metrolist.music.extensions.isSyncEnabled
+import com.metrolist.music.utils.isSignedInToYouTube
 import com.metrolist.music.utils.rememberPreference
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -51,7 +52,7 @@ fun CreatePlaylistDialog(
     val context = LocalContext.current
 
     val innerTubeCookie by rememberPreference(InnerTubeCookieKey, "")
-    val isSignedIn = innerTubeCookie.isNotEmpty()
+    val isSignedIn = isSignedInToYouTube(innerTubeCookie)
 
     val notLoggedInYoutubeStr = stringResource(R.string.not_logged_in_youtube)
     val syncDisabledStr = stringResource(R.string.sync_disabled)
--- a/app/src/main/kotlin/com/metrolist/music/ui/component/CreatePlaylistDialog.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/component/CreatePlaylistDialog.kt
@@ -27,8 +27,7 @@ import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.input.TextFieldValue
 import androidx.compose.ui.unit.dp
-import com.metrolist.innertube.YouTube
-import com.metrolist.music.LocalDatabase
+import com.metrolist.music.LocalSyncUtils
 import com.metrolist.music.R
 import com.metrolist.music.constants.InnerTubeCookieKey
 import com.metrolist.music.db.entities.PlaylistEntity
@@ -38,7 +37,6 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import java.time.LocalDateTime
-import java.util.logging.Logger
 
 @Composable
 fun CreatePlaylistDialog(
@@ -47,7 +45,7 @@ fun CreatePlaylistDialog(
     allowSyncing: Boolean = true,
     onPlaylistCreated: ((String) -> Unit)? = null,
 ) {
-    val database = LocalDatabase.current
+    val syncUtils = LocalSyncUtils.current
     val coroutineScope = rememberCoroutineScope()
     var syncedPlaylist by remember { mutableStateOf(false) }
     val context = LocalContext.current
@@ -64,33 +62,18 @@ fun CreatePlaylistDialog(
         initialTextFieldValue = TextFieldValue(initialTextFieldValue ?: ""),
         onDismiss = onDismiss,
         onDone = { playlistName ->
-            coroutineScope.launch(Dispatchers.IO) {
-                val browseId =
-                    if (syncedPlaylist && isSignedIn) {
-                        YouTube.createPlaylist(playlistName)
-                    } else if (syncedPlaylist) {
-                        Logger.getLogger("CreatePlaylistDialog").warning("Not signed in")
-                        return@launch
-                    } else {
-                        null
-                    }
-
-                val playlistEntity =
-                    PlaylistEntity(
-                        name = playlistName,
-                        browseId = browseId,
-                        bookmarkedAt = LocalDateTime.now(),
-                        isEditable = true,
-                    )
-
-                database.query {
-                    insert(playlistEntity)
-                }
-
-                withContext(Dispatchers.Main) {
-                    onPlaylistCreated?.invoke(playlistEntity.id)
-                }
-            }
+            // The dialog is already dismissed by the time this runs, so the work is handed to
+            // SyncUtils and its application-lifetime scope: the playlist is written locally first
+            // and only then offered to YouTube, which may fail or be unreachable.
+            syncUtils.createPlaylist(
+                PlaylistEntity(
+                    name = playlistName,
+                    bookmarkedAt = LocalDateTime.now(),
+                    isEditable = true,
+                    isAutoSync = syncedPlaylist && isSignedIn,
+                ),
+                onCreated = onPlaylistCreated,
+            )
         },
         extraContent = {
             if (allowSyncing) {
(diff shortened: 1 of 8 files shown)


makro17:
The finding is correct as stated: `withRetry` around `createPlaylist` can produce a second remote playlist if YouTube creates one and the response never arrives. `playlistsBeingCreated` guards concurrent attempts, not sequential retries, exactly as noted.

One piece of context that I think changes what the right fix is. `InnerTube.createPlaylist` already wraps its own POST in a `withRetry` of its own (`InnerTube.kt:580-593`, retrying up to 3 times on `IOException`), and a response lost after the server committed is precisely an `IOException`. So the non-idempotent retry is not introduced here — it is on `main` today, one layer down. What this pull request adds is a second retry loop around it, which turns a worst case of 3 create requests into 9.

Given that, reconciling before retrying looks like the wrong shape of fix:

- It means listing the account's playlists and matching by name before each attempt. Name is not a key: a user who genuinely wants two playlists called "Rock" would have the second creation swallowed. That trades a visible duplicate for silent data loss, which is worse.
- `CreatePlaylistBody` carries only `context` and `title`, so there is no client-supplied identifier to match on instead. If the endpoint accepts an idempotency parameter that I have not found, I would rather use that than match on name.

The cheap fix that actually helps is to **drop the outer `withRetry` here**, leaving InnerTube's, so this path is no worse than `main` and the worst case goes back to 3. The retry loop is not needed for correctness anyway: on failure the playlist stays pending, and the next sync pass is the real retry — that is the mechanism this pull request is built on, and it was verified on device across three sync passes and a cold restart, including one after the app was killed.

Happy to push that. If a maintainer would rather see reconciliation, or wants the inner retry addressed too since it is the pre-existing half, say so and I will do that instead.

Worth recording either way: this affects only the creation path. Once `browseId` is written back, the `isAutoSync = true AND browseId IS NULL` filter stops matching, so no later sync re-creates the playlist.

nyxiereal:
The local-first behavior is valuable, but `createPlaylist` must not be blindly retried: an ambiguous timeout after remote success can create duplicate playlists. Please make creation single-attempt with explicit pending/reconciliation behavior, use the same `SAPISID`-based login predicate as sync, and test lost responses plus partial-cookie sessions.

makro17:
Done, rebased onto `9285e7a96` and pushed as `2a5853382`. Taking the three in the order they matter.

**One thing to look at before the rest, because it changes behaviour that is already on `main`.** Making creation single-attempt is not something this pull request can do on its own: `InnerTube.createPlaylist` wraps its own POST in `withRetry`, up to three attempts on `IOException`, and a response lost after YouTube committed the playlist *is* an `IOException`. With the outer retry this pull request had added, the worst case was nine requests for one playlist. Both are gone now, so it is one request. That means a one-line change in `InnerTube.kt`, outside this pull request's original blast radius, and if you would rather keep the retry there and have me solve it another way, say so and I will.

The reasoning for removing it rather than working around it: a creation carries no key, so a retry cannot be made safe at that layer — it can only be made lucky. Every other endpoint under `withRetry` is either a read or an edit that names the thing it edits.

**Reconciliation.** This also answers the objection I raised myself last time, that matching a lost creation by name trades a visible duplicate for silent data loss, because a user is entitled to a second playlist called "Rock". What identifies the playlist is not its name but the moment it appeared. The account's playlists are read before the request goes out; if the request fails, they are read again. A playlist that was not there before, is there now, and carries the name that was asked for is the one that request created, and the local row claims it. Nothing new, or more than one new, leaves the playlist pending for the next sync rather than guessing.

The read before the attempt is taken once per batch of pending playlists rather than once each, and the read after only happens when something failed, so the path where everything works costs one extra request per creation and nothing else.

**The login predicate.** `CreatePlaylistDialog` asked `innerTubeCookie.isNotEmpty()` while `isLoggedIn()` requires `SAPISID` in the parsed cookie. A session without `SAPISID` cannot sign a write, so a playlist that dialog marked for syncing could never reach YouTube and would stay pending forever, which is precisely the state this pull request is built around. Both now go through one `isSignedInToYouTube`, which also stops treating a cookie that fails to parse as a session.

**Tests**, twelve of them, on the two cases you named:

- Lost response: the playlist that appeared is claimed; one that already carried the name is not, since it is not evidence this request landed; two new ones under the name claim neither; the same one listed twice is still claimed; the name is matched exactly.
- Partial cookie: with `SAPISID` yes, without it no, empty no, unparseable no, and `__Secure-3PAPISIDTS` does not pass as `SAPISID`.

`assembleFossDebug` and 99 tests with none failing.

One hole I went looking for and did not find: there is no "succeeded but returned no id" path that would leave a playlist pending without reconciling. `CreatePlaylistResponse.playlistId` is non-null and required, so a response arriving without it fails to deserialise and lands in the failure branch, where the reconciliation picks it up.

makro17:
Rebased onto `9c991782e` and pushed as `66d46bba9`. One correction to what I wrote yesterday, in your favour.

I said this needed a one-line change in `InnerTube.kt` to drop the `withRetry` around `createPlaylist`, and offered to solve it another way if you would rather keep it. That is no longer mine to do: the InnerTubeX move already took `createPlaylist` out of `withRetry`, which now wraps only `returnYouTubeDislike`. So the change is gone from this pull request and creation is single-attempt without touching `innertube` at all. The reconciliation and the `SAPISID` predicate are unchanged.

Also worth knowing before you look at the build: `main` does not compile as it stands, so neither does anything based on it. The `innertubex` pin is one release behind the methods `InnerTube.kt` calls. It is #4291, one line, with the details there. I ran this pull request's build with that fix applied locally: `assembleFossDebug` and 100 tests, none failing.

nyxiereal:
Superseded by 5247e9a39.

PR #4272, opened 2026-08-17
fix(sync): make playlist additions reach YouTube reliably

makro17:
## Problem

Songs added to a local playlist do not reliably reach the mirrored YouTube playlist, especially when
several are added at once. A later sync, seeing them only locally, then deletes them from the
device — so the songs are lost in both places.

Separately, adding an **album** to a synced playlist puts every song on YouTube **twice**.

## Cause

Three defects compound.

**1. Fire-and-forget uploads, at eight call sites.** `SongMenu`, `SelectionSongsMenu`, `PlayerMenu`,
`QueueMenu` and `YouTubeSongMenu` each fire an independent `YouTube.addToPlaylist` from inside their
`onGetSong` callback — whose documented job is only to insert song rows — in parallel with the
dialog's own upload, bypassing the edit throttle and discarding the result.

`AlbumMenu`, `YouTubeAlbumMenu` and `YouTubePlaylistMenu` do the same with `addPlaylistToPlaylist`,
over the very id set the dialog then uploads one at a time. That is the duplicate-album case above:
one bulk request plus one per-song loop, over the same songs.

**2. Uploads die with the dialog.** `AddToPlaylistDialog` writes the local rows first, then uploads
song by song on `rememberCoroutineScope()`. Every upload passes through the 500 ms playlist-edit
throttle, so thirty songs take at least fifteen seconds. When the host menu leaves composition the
scope is cancelled and the remaining uploads never happen — but the local rows are already
committed. That mismatch is exactly what the next sync deletes.

**3. `setVideoId` is never recorded.** `YouTube.addToPlaylist` returns the raw response without
parsing it, so the app cannot tell a row that YouTube accepted from one it never saw.

## Solution

**Delete all eight fire-and-forget uploads**, leaving the dialog's single upload path. The `insert`
calls stay. Removing the bulk album request costs a fifty-song album roughly twenty-five seconds of
throttled per-song uploads in the background instead of one immediate request; that is accepted,
because the uploads now survive dismissal so their duration stops being the user's problem, and the
throttle exists because YouTube rate-limits playlist edits. Album songs gain `setVideoId`
persistence as a by-product.

**Upload on `syncScope`.** New `SyncUtils.addSongsToPlaylist(browseId, playlistId, playlistName,
songIds)` launches on the application-lifetime `SupervisorJob` scope instead of the caller's
composition scope, mirroring its neighbour `scheduleRemoveFromPlaylist`. It processes ids
sequentially through the existing throttle and wraps each in `withRetry`. The suspending single-song
`addToPlaylist` is deleted; the dialog was its only caller.

`withRetry` goes **outside** `runQueuedPlaylistEdit`, not inside: inverted, the 1 s and 2 s backoffs
would elapse while holding `playlistEditMutex` and stall every playlist edit in the app for up to
three seconds. `runQueuedPlaylistEdit` becomes generic so the block can return the parsed
`setVideoId`; its four existing callers infer `Unit` and are untouched.

The batch is marked with `markPlaylistModifying` once for the whole list rather than once per song,
which incidentally closes a real gap: today the playlist is unmarked between consecutive songs, so a
sync can interleave with an upload in progress.

Because uploads now outlive the dialog, the dialog returns to the obvious order: local writes,
uploads, dismiss.

**Parse and persist `setVideoId`.** The response model already existed with the right shape and was
unused. Since it had never been deserialised against a real response its non-null fields were a
hazard — a missing `playlistEditResults` would throw, `runCatching` would report failure for a song
that was in fact added, and the retry would duplicate it on YouTube — so that field defaults to
empty and `status` becomes nullable. Persistence uses one new query and **no schema change**; its
subselect fills the first unconfirmed row, which is what the "add anyway" case needs.

Two benefits: `scheduleRemoveFromPlaylist` no longer has to refetch an entire playlist from YouTube
to discover a `setVideoId`, and `setVideoId IS NULL` becomes a meaningful "not confirmed on YouTube"
signal.

**Report failures.** Uploads that fail after all retries raise a `Toast` naming the playlist instead
of only reaching `Timber.e`. One new plural string.

Deliberately out of scope: uploads do not survive the process being killed. Making them durable
would need either WorkManager, which is not a dependency here, or a persisted queue table, which
would be a schema change.

## Testing

- `./gradlew :app:assembleFossDebug`, `./gradlew :app:testFossDebugUnitTest` and
  `./gradlew :innertube:testDebugUnitTest`.
- The hardened response model is covered by unit tests, including a response with
  `playlistEditResults` absent, which is the case that used to turn a successful add into a retry.
- On device, signed in, against a disposable playlist:
  - seven songs added from `SelectionSongsMenu` with the menu dismissed immediately — the dialog was
    gone two seconds in, the throttled batch needs three and a half, and all seven rows ended with a
    non-null `setVideoId`, which can only come from YouTube's reply. Every upload landed after the
    scope that used to own it was gone.
  - the same playlist reported 7 remote songs, not 14, which is the duplicate upload removed.
  - with the network down, one upload logged three attempts at 1 s and 2 s backoffs and then
    reported "1 song was not added to YouTube" rather than failing silently.

## Summary by CodeRabbit

* **New Features**
  * Added batch playlist uploads for faster multi-song additions.
  * Added retry handling and clearer notifications when songs fail to upload.
  * Playlist additions now sync returned video identifiers automatically.
  * Added artist-name alias support and improved multi-artist editing.
  * Improved handling of incomplete playlist responses and duplicate entries.

* **Bug Fixes**
  * Ensured playlist changes finish syncing before dialogs close.
  * Improved library synchronization while preserving downloaded songs and artist mappings.

* **Tests**
  * Added coverage for playlist upload reconciliation, response parsing, and duplicate entries.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/db/DatabaseDao.kt
+++ b/app/src/main/kotlin/com/metrolist/music/db/DatabaseDao.kt
@@ -1240,20 +1240,20 @@ interface DatabaseDao {
     @Query("UPDATE playlist_song_map SET position = position + :delta WHERE playlistId = :playlistId")
     fun shiftPlaylistSongPositions(playlistId: String, delta: Int)
 
-    // The subselect matters: adding a song that is already in the playlist leaves two rows for it,
-    // and only the one still awaiting confirmation should take the new setVideoId.
+    // By row id, not by (playlistId, songId): a playlist may hold the same song twice, and the
+    // setVideoId names one of those two items on YouTube rather than the song.
+    @Query("UPDATE playlist_song_map SET setVideoId = :setVideoId WHERE id = :mapId")
+    fun updatePlaylistSongSetVideoId(mapId: Long, setVideoId: String)
+
+    // The copies of this song the playlist has already had confirmed, which is what an upload
+    // whose answer went missing is weighed against.
     @Query(
         """
-        UPDATE playlist_song_map SET setVideoId = :setVideoId
-        WHERE id = (
-            SELECT id FROM playlist_song_map
-            WHERE playlistId = :playlistId AND songId = :songId AND setVideoId IS NULL
-            ORDER BY position
-            LIMIT 1
-        )
+        SELECT setVideoId FROM playlist_song_map
+        WHERE playlistId = :playlistId AND songId = :songId AND setVideoId IS NOT NULL
         """
     )
-    fun updatePlaylistSongSetVideoId(playlistId: String, songId: String, setVideoId: String)
+    fun playlistSongSetVideoIds(playlistId: String, songId: String): List<String>
 
     @Transaction
     fun addSongToPlaylist(playlist: Playlist, songIds: List<String>) {
@@ -1274,54 +1274,44 @@ interface DatabaseDao {
     }
 
     // This prevents songs from being removed during automatic playlist synchronization
+    /**
+     * Returns the row each song was inserted as, paired with its id, so a caller uploading them
+     * can later say which row an answer belongs to. A playlist is allowed to hold the same song
+     * twice, and the two rows are only told apart by their id.
+     */
     @Transaction
     fun addSongsToPlaylist(
         playlist: Playlist,
         songs: List<Pair<String, String?>>, // Pair of (songId, setVideoId)
         prepend: Boolean = false,
-    ) {
+    ): List<Pair<String, Long>> {
         val now = LocalDateTime.now()
         val songsToInsert =
             songs.mapNotNull { (id, setVideoId) ->
                 getSongByIdBlocking(id)?.let { id to setVideoId }
             }
-        if (songsToInsert.isEmpty()) return
+        if (songsToInsert.isEmpty()) return emptyList()
 
         if (prepend) {
             shiftPlaylistSongPositions(playlist.id, songsToInsert.size)
-            var position = 0
-            songsToInsert.forEach { (id, setVideoId) ->
-                val existingSong = getSongByIdBlocking(id)!!
-                if (existingSong.song.inLibrary == null) {
-                    inLibrary(id, now)
-                }
-                insert(
-                    PlaylistSongMap(
-                        songId = id,
-                        playlistId = playlist.id,
-                        position = position++,
-                        setVideoId = setVideoId,
-                    ),
-                )
+        }
+        var position = if (prepend) 0 else playlist.songCount
+        val inserted = songsToInsert.map { (id, setVideoId) ->
+            val existingSong = getSongByIdBlocking(id)!!
+            if (existingSong.song.inLibrary == null) {
+                inLibrary(id, now)
             }
-        } else {
-            var position = playlist.songCount
-            songsToInsert.forEach { (id, setVideoId) ->
-                val existingSong = getSongByIdBlocking(id)!!
-                if (existingSong.song.inLibrary == null) {
-                    inLibrary(id, now)
-                }
-                insert(
-                    PlaylistSongMap(
-                        songId = id,
-                        playlistId = playlist.id,
-                        position = position++,
-                        setVideoId = setVideoId,
-                    ),
-                )
-            }
+            id to insert(
+                PlaylistSongMap(
+                    songId = id,
+                    playlistId = playlist.id,
+                    position = position++,
+                    setVideoId = setVideoId,
+                ),
+            )
         }
         updatePlaylistLastUpdated(playlist.id)
+        return inserted
     }
 
     fun downloadedSongs(
@@ -1804,7 +1794,7 @@ interface DatabaseDao {
     fun insert(map: AlbumArtistMap)
 
     @Insert(onConflict = OnConflictStrategy.IGNORE)
-    fun insert(map: PlaylistSongMap)
+    fun insert(map: PlaylistSongMap): Long
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     fun insert(searchHistory: SearchHistory)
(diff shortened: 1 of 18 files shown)


makro17:
Three findings, and they do not all weigh the same. Taking them in order of what I think matters.

**Retrying `addToPlaylist` is not idempotent — correct, with context.** As with the sibling finding on #4271, `InnerTube.addToPlaylist` already wraps its POST in its own `withRetry` (`InnerTube.kt:496-500`, up to 3 attempts on `IOException`), and a response lost after the edit committed is exactly an `IOException`. So the hazard is on `main` today, one layer down; what this pull request adds is an outer loop that raises the worst case from 3 requests to 9.

Reconciling the playlist before each retry means fetching it and checking whether the song arrived, which is a request per retry against the endpoint we are already being throttled on, and it cannot distinguish "my upload landed" from "the same song was already there" — that state is legitimate, since the duplicate warning's "add anyway" exists precisely to allow it.

There is a cheaper mitigation specific to this path, and it is already half-built: the response carries a `setVideoId`, and the change persists it. A retry that finds the row already holds one has evidence the previous attempt landed and can stop. That is a small change and I am happy to push it. Dropping the outer `withRetry`, as I have proposed on #4271, is the other option and brings this back to `main`'s behaviour.

**Targeting the exact inserted row rather than the first unconfirmed one.** I would rather not, and here is the reasoning to argue against. The subselect deliberately fills the first row with a null `setVideoId` for that song in that playlist. When two rows exist for the same song in the same playlist, they are interchangeable: each `setVideoId` names one of two identical items on YouTube, and removing either row removes one of them. Crossing the two values changes which row names which item and nothing the user can observe. Threading the inserted row id through would mean changing `addSongsToPlaylist`'s signature and the DAO insert path to return ids, for an outcome that is equivalent. If there is a case where the two are not interchangeable, I would like to see it — that would change my mind.

**The response model.** This one I would like to keep as it is, and it is worth saying why, since it looks lax: `playlistEditResults` defaults to empty and `status` is nullable on purpose. The model had never been deserialised against a real response before this change. Had a required field been absent, `runCatching` would have reported a failure for a song that YouTube had in fact added, and the retry would have duplicated it — the very outcome the first finding is about. Being lenient here is what keeps a successful add from being retried.

Summary of what I will push if there is no objection: stop the retry once the row has a `setVideoId`, and drop the outer `withRetry` in the sibling pull request. I would leave the row targeting and the response model alone. Tell me if you would rather have it the other way and I will follow.

nyxiereal:
Please address row identity and retry ambiguity before merge. Updating the first pending `(playlistId, songId)` row can attach `setVideoId` to the wrong duplicate occurrence, and blindly retrying `addToPlaylist` can duplicate a remote item when the first request succeeded but its response was lost. Use stable occurrence identity, reconcile ambiguous results instead of replaying the mutation, make incomplete nested responses safe to decode, and cover duplicates/lost responses in tests.

makro17:
All four addressed, rebased onto `9c991782e` and pushed as `18a43b2ae`.

**Row identity.** The setVideoId was written to the first row for that song still lacking one, picked by position. `addSongsToPlaylist` now returns the row it created for each song, paired with its id, and the upload updates that row by id. This turned out to be far smaller than I feared when I argued against it: adding a return value does not disturb the callers that ignore it, so only the one caller that needs the ids changed. That was the weakest part of my earlier reasoning, and it was wrong of me to weigh the cost without checking it.

The argument I made for why the two rows were interchangeable does not survive either, now that I have written the test: two rows for the same song each carry their own setVideoId, and `PlaylistOccurrenceTest` shows confirming one leaves the other untouched. Under the old query the second upload could not reach the second row at all, since the subselect always found the first unconfirmed one.

**Reconciling instead of replaying.** Your refactor took care of half of this before I got to it: `InnerTube.addToPlaylist` no longer goes through `withRetry`, so the layer I was going to have to argue about is already single-attempt. The outer retry this pull request had added is gone too.

What replaces it is a comparison rather than a repeat. Asking "is the song in the playlist" answers the wrong question, because a playlist is allowed to hold the same song twice and "add anyway" exists to put it there, so presence proves nothing. Counting does prove it: the copies of that song the playlist holds now, against the ones this device had already confirmed. One more than before means the request landed, and the copy that was not there before is the one it created. Only when the playlist says nothing arrived is the upload sent again, and by then it cannot duplicate anything.

The copies this device already knows about come from the local rows, so the happy path costs no extra request; the playlist is only read when an upload has already failed.

**The response model.** You were right that this needed fixing and I was wrong to defend it as it stood. I had made the two outer fields optional for exactly the reason you gave — a required field missing turns a successful add into a failure, and a failure is what makes the caller ask again — and then left the nested ones required, which is the same trap one level down. All of them are optional now, and the setVideoId is taken from the first result that carries one rather than from the first result.

**Tests**, seventeen across the three:

- Reconciliation, eight: a copy that was not there before, a playlist holding no more than before, a song already present not mistaken for this upload, a first copy, a second copy of a song already held twice, two copies appearing at once landing without naming either, a copy this device never recorded, and a playlist that lost copies.
- Occurrence identity, three, against a real database: two rows for one song get distinct ids, confirming one leaves the other null, and both can be confirmed separately.
- Decoding, six: an edit result with no added data, added data with no ids, an empty body, and a setVideoId later in the results.

`assembleFossDebug`, 94 app tests and 16 innertube tests, none failing.

One thing you should know about how that was verified: `main` does not compile as it stands, and neither did any branch on top of it. The `innertubex` pin is a release behind the methods `InnerTube.kt` calls. I raised it separately as #4291 with the details rather than smuggle a dependency change in here, and ran the build for this pull request with that one-line fix applied locally.

nyxiereal:
Superseded by 5247e9a39.

PR #4273, opened 2026-08-17
fix(sync): stop discarding songs a playlist only has locally

makro17:
## Problem

Syncing a playlist deletes songs the device holds and YouTube does not, silently and without asking.

Any song that failed to upload — a dropped request, no network at the time, the app closed
mid-upload — is removed from the device on the next sync. The user never sees it happen, and the
song is gone from both sides.

On a real library this is not a corner case. In the library this was tested against, one playlist
held 37 songs of which YouTube listed 7: a playlist can be almost entirely local while presenting
itself as synced.

## Cause

`executeSyncPlaylist` calls `clearPlaylist(playlistId)` and repopulates from the remote playlist.
YouTube is treated as absolute truth. There is already a mechanism that carries songs across that
clear-and-repopulate cycle, but it preserves **downloaded** songs only, so everything else that is
absent remotely is dropped.

**Fixing that one function is not enough.** The **Sync** item in a playlist's own menu, shown for
every playlist with a `browseId`, never went through `SyncUtils` at all: its handler in
`LocalPlaylistScreen` fetched the remote page and ran its own `clearPlaylist` and repopulate. One
tap deleted the local-only songs even with the fix above in place, and being a private copy it also
missed the edit throttle, the modification guard and the retries.

## Solution

Generalise the existing preservation from "downloaded songs" to "all songs present locally but
absent remotely", which are appended after the remote ordering. Nothing else about the sync changes.

Route the playlist menu's **Sync** through the same path: the handler now calls a new
`SyncUtils.syncPlaylistSuspend`, added beside the other suspend entry points, and the duplicated
copy in the screen is deleted. Nine lines shorter, and the screen inherits the preservation instead
of reimplementing the destruction.

Accepted trade-off: a song deleted from the YouTube side no longer disappears from the device
automatically. Choosing between "delete the user's song without asking" and "keep a song the user
may have removed elsewhere", this errs toward not destroying data.

Recorded for whoever reads this later: `SyncOperation.SinglePlaylist` still has no producer. It was
declared for exactly this action and never wired up, which is how the screen came to carry its own
copy in the first place.

## Testing

- `./gradlew :app:assembleFossDebug` and `./gradlew :app:testFossDebugUnitTest`.
- On device, against a restored real library of 4318 playlist rows across 22 synced playlists:
  - a full sync preserved exactly 42 songs across 5 playlists, which is precisely the set an
    independent comparison of the user's backup against YouTube had identified beforehand. None
    were lost, and all were appended after the remote ordering.
  - tapping **Sync** in the playlist menu of a playlist holding 37 songs of which YouTube lists 7
    logged "Preserved 30 songs absent from the remote playlist" and left 37 rows at contiguous
    positions 0–36, where before the same tap deleted 30 of them.
- Songs that are present on both sides are untouched, and playlists with no local-only songs sync
  exactly as before.

## Summary by CodeRabbit

* **New Features**
  * Playlist synchronization now retains locally saved songs that are no longer present remotely.
  * Preserved songs remain in their original order after current remote entries.
  * Artist subscriptions use resolved artist names for improved accuracy.

* **Bug Fixes**
  * Incomplete playlist reads no longer overwrite local playlists.
  * Missing artist information is repaired during synchronization.
  * Library cleanup better preserves downloaded content and removes outdated entries.
  * Synchronization and playlist edits now run more consistently.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/ui/screens/playlist/LocalPlaylistScreen.kt
+++ b/app/src/main/kotlin/com/metrolist/music/ui/screens/playlist/LocalPlaylistScreen.kt
@@ -104,7 +104,6 @@ import androidx.navigation.NavController
 import coil3.compose.AsyncImage
 import com.metrolist.innertube.YouTube
 import com.metrolist.innertube.models.PlaylistItem
-import com.metrolist.innertube.models.SongItem
 import com.metrolist.innertube.utils.completed
 import com.metrolist.music.LocalDatabase
 import com.metrolist.music.LocalDownloadUtil
@@ -124,7 +123,6 @@ import com.metrolist.music.db.entities.PlaylistSong
 import com.metrolist.music.db.entities.PlaylistSongMap
 import com.metrolist.music.extensions.move
 import com.metrolist.music.extensions.toMediaItem
-import com.metrolist.music.models.toMediaMetadata
 import com.metrolist.music.playback.ExoDownloadService
 import com.metrolist.music.playback.queues.ListQueue
 import com.metrolist.music.ui.component.ActionPromptDialog
@@ -1390,19 +1388,10 @@ fun LocalPlaylistHeader(
                             onEdit = onShowEditDialog,
                             onSync = {
                                 scope.launch(Dispatchers.IO) {
-                                    val playlistPage =
-                                        YouTube
-                                            .playlist(playlist.playlist.browseId!!)
-                                            .completed()
-                                            .getOrNull() ?: return@launch
-                                    database.transaction {
-                                        clearPlaylist(playlist.id)
-                                        val songIds = playlistPage.songs
-                                            .map(SongItem::toMediaMetadata)
-                                            .onEach(::insert)
-                                            .map { it.id to it.setVideoId }
-                                        addSongsToPlaylist(playlist, songIds)
-                                    }
+                                    syncUtils.syncPlaylistSuspend(
+                                        playlist.playlist.browseId!!,
+                                        playlist.id,
+                                    )
                                     withContext(Dispatchers.Main) {
                                         snackbarHostState.showSnackbar(playlistSyncedStr)
                                     }
--- a/app/src/main/kotlin/com/metrolist/music/utils/PlaylistPreservation.kt
+++ b/app/src/main/kotlin/com/metrolist/music/utils/PlaylistPreservation.kt
@@ -33,3 +33,27 @@ fun songIdsAbsentFromRemote(
         }
     }
 }
+
+/**
+ * The number of songs a remote playlist says it holds, read out of the localised text YouTube
+ * renders under its title, such as "12 songs". Null when the page says nothing countable.
+ */
+fun remoteSongCountOf(songCountText: String?): Int? =
+    songCountText?.let { Regex("""\d+""").find(it)?.value?.toIntOrNull() }
+
+/**
+ * Whether a remote read describes the playlist well enough to rebuild the local copy from it.
+ *
+ * [readSongCount] is how many songs came through and [claimedSongCount] is how many the page says
+ * are there. A read that falls short of the claim is one this build could not follow to the end,
+ * and it is not the same thing as a playlist that lost songs: every song that failed to arrive
+ * would be taken for local-only and moved to the end of the playlist. An empty read is the case
+ * that matters most, because nothing at all arrived.
+ *
+ * A page that claims a number no larger than what arrived is taken at its word, and that includes
+ * a playlist which genuinely holds nothing. A page that claims nothing countable is only trusted
+ * when at least one song arrived, since there is then no way to tell an empty playlist from an
+ * unreadable one.
+ */
+fun remoteReadAccountsForPlaylist(readSongCount: Int, claimedSongCount: Int?): Boolean =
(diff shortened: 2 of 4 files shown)


makro17:
Good catch, and fixed in bc43761b8.

The scenario is reachable rather than theoretical: the duplicate warning's "add anyway" deliberately leaves a second row for the same song in the same playlist, so a playlist holding a song twice while YouTube lists it once is a state the app produces on purpose. The rebuild then kept only the remote copy and dropped the other, which is exactly the kind of loss this pull request exists to stop.

Comparison is now by occurrence: each local row is matched against one remote listing, and the rows left unmatched are the ones preserved. I extracted it as a pure function, `songIdsAbsentFromRemote`, so the rule is covered on the JVM rather than only in the sync — nine cases, including `[a, a]` against `[a]`, the reverse, and preservation order.

One thing I did not change: the `filterNot { it.id in remoteIds }` calls elsewhere in this file, for library songs, albums, artists and podcast episodes. Membership is the right comparison there, since an item is either in the library or it is not — only playlists can hold the same entry more than once.

nyxiereal:
Two cases still need fixing before this is safe to merge: a successful empty remote playlist currently hits the existing early return, and `syncPlaylistSuspend()` can race queued local playlist edits. Please distinguish fetch failure from a valid empty result, reconcile the empty case, serialize manual sync with local modifications, and add tests for both paths.

makro17:
Both findings hold up, and the empty case turned out to be worth more than the empty playlist it is named after. Pushed as `cb68a9fc4`, on top of a rebase onto current `main`.

**Telling a valid empty result from a read that failed.** The skip lived inside the success branch, so a page this build could not follow to the end and a playlist that genuinely holds nothing left through the same door. The signal to separate them was already in the file: `executeSyncSavedPlaylists` reads the song count out of `songCountText` and stores it as `remoteSongCount`, and `update(playlistEntity, playlistItem)` does the same. A read is now weighed against that count.

That covers more than the empty case, which is why I would rather have it this way round. A partial read — 3 songs of 300 because pagination broke — is the same class of failure and worse in its effect: with this pull request's preservation rule, the 297 that never arrived are all unaccounted for remotely, so they would be preserved and appended after the 3 that did. The playlist survives, but the user's ordering does not. The check leaves the playlist alone until a read adds up.

The rule is two pure functions, `remoteSongCountOf` and `remoteReadAccountsForPlaylist`, with nine cases: a full read, a short read, an empty read against a page claiming songs, an empty read against a page claiming none, an empty read against a page claiming nothing countable, a read longer than the claim (the count under the title can lag an edit that already landed), and the parse itself. A page that claims nothing countable is trusted only when at least one song arrived, since there is otherwise no way to tell an empty playlist from an unreadable one — that keeps today's conservative behaviour for that case.

**Reconciling the empty case.** Worth stating plainly, because the outcome is quieter than it sounds: under the preservation rule a valid empty remote accounts for none of the local songs, so reconciling keeps every one of them. The change that reconciling does make is that the `setVideoId` on each row is cleared, and that is correct — a `setVideoId` names an item in the remote playlist, and the remote no longer lists it. So the early return goes and the empty result flows through the normal path. If you meant something stronger by reconcile, say so and I will follow.

**Serializing the manual sync.** `syncPlaylistSuspend` now runs inside `runQueuedPlaylistEdit`, so it queues behind an upload or a removal for that playlist instead of reading the remote while one is in flight and rebuilding from an answer that edit has not reached. I checked the lock order before doing it: nothing takes `playlistEditMutex` before `syncExecutionMutex`, so there is no inversion, and the throttle it waits on is 500 ms.

**Testing.** `assembleFossDebug`, and the suite at 100 tests with none failing. The empty path's outcome is covered at the rule level by `everything is preserved when the remote lists nothing`, which is exactly that case.

One gap I would rather name than paper over: the serialization itself has no unit test. `SyncUtils` takes a `Context` and the database by injection, so covering it on the JVM means extracting the ordering into something testable. Happy to do that if you want it covered rather than argued.

Unrelated, noticed while reading: `SyncOperation.SinglePlaylist` is declared and dispatched in `processOperation`, but nothing enqueues it. Left alone, since it is not this pull request's business.

makro17:
Rebased again onto `9285e7a96`, which landed while I was writing the previous comment and rewrote the same part of `executeSyncPlaylist`. Now at `9a7086e7b`, merging cleanly. What the rebase changed in this pull request, so the diff is not a surprise:

- The preserved set is read from `localIds`, which `9285e7a96` already loads through `playlistSongIds` and which is ordered by position and keeps duplicates. That is exactly what the occurrence comparison needs, so the extra `playlistSongs(playlistId).first()` this pull request used to do is gone.
- The `downloadedSongIds` block is still what this change replaces, and the replacement is a superset: a downloaded song the remote does not list is a local row the remote does not account for, so it is preserved either way, along with every song that is not downloaded.
- One consequence worth flagging rather than deciding on my own: `downloadedPlaylistSongIds`, added in `9285e7a96`, has no caller left after this. I have not touched it, since removing a query you added hours ago is not mine to do. Say the word and I will drop it, or leave it.
- The metadata repair that `9285e7a96` added, `metadataInserts` and `songIdsWithoutArtists`, is untouched and still runs before the preserved songs are appended.

Rebuilt and retested on top of it: `assembleFossDebug`, and 100 tests with none failing.

nyxiereal:
Superseded by 5247e9a39.

PR #4347, opened 2026-09-04
feat: add Zemer lyrics provider

alltechdev:
Adds a lyrics provider backed by [Zemer](https://search.zemer.io), a catalog of Jewish music lyrics.

Unlike title-search providers, it resolves by the exact YouTube video ID: `/lyrics/resolve?videoId=` returns ranked source pointers (synced karaoke feeds, LrcLib track IDs, lyric-site pages, inline text), and the provider fetches them in rank order, preferring synced, returning the first body with at least 4 lines. A track outside the catalog is a single fast 404, so the chain is never delayed.

Follows the existing provider convention: `LyricsProvider` object, dataStore toggle (default on), entry in the registry and the settings enable/priority dialogs. Placed after the general-purpose providers in the default order.

Built with `:app:assembleFossDebug` and tested on a device: synced and plain lyrics both render, and out-of-catalog tracks fall through to the next provider immediately.

## Summary by CodeRabbit

* **New Features**
  * Added Zemer as a lyrics provider, offering lyrics for Jewish music.
  * Zemer supports multiple lyric sources and formats, including synchronized lyrics.
  * Newly added providers are automatically included in saved provider lists while preserving existing ordering.
* **Settings**
  * Added “Enable Zemer” with a description explaining its lyric coverage.
  * Zemer can be enabled or disabled and placed within the preferred provider order.

### DIFF
--- a/app/src/main/kotlin/com/metrolist/music/constants/PreferenceKeys.kt
+++ b/app/src/main/kotlin/com/metrolist/music/constants/PreferenceKeys.kt
@@ -81,6 +81,7 @@ const val SYSTEM_DEFAULT = "SYSTEM_DEFAULT"
 val AppLanguageKey = stringPreferencesKey("appLanguage")
 val ContentLanguageKey = stringPreferencesKey("contentLanguage")
 val ContentCountryKey = stringPreferencesKey("contentCountry")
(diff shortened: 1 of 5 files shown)


nyxiereal:
LGTM

PR #3446, opened 2026-04-04
fix(playlist): ensure songs persist across sync by marking them as inLibrary

kairosci:
## Problem
Songs added to playlists from any menu entry point were being removed during automatic synchronization. This occurred because songs added to playlists without the `inLibrary` marker were not preserved across sync operations, causing them to disappear unpredictably.

## Cause
The application had multiple entry points for adding songs to playlists, but only some were using the proper `addSongToPlaylistWithLibrarySync()` function which marks songs as belonging to the library. Additionally, the implementation had critical issues:
- Lost `setVideoId` metadata needed for reordering operations
- Race conditions in async playlist creation flows
- Non-atomic transactions causing data inconsistency
- Synchronization paths that truncated playlists on cancellation

## Solution
Implemented a complete structural fix across all 12 entry points with comprehensive robustness improvements:

### 1. Enhanced addSongToPlaylistWithLibrarySync()
- Changed signature to accept `List<Pair<String, String?>>` (songId, setVideoId)
- Preserves `setVideoId` in PlaylistSongMap for YouTube reordering operations
- Maintains library sync marker functionality

### 2. Atomic Transactions
- **HomeScreen.kt**: Single `withTransaction` wrapping playlist insert + songs + linking (line 444-461)
- **OnlinePlaylistScreen.kt**: First-time save consolidated into one transaction (line 588-615)
- **SyncUtils.kt**: Playlist clear + repopulate now atomic (line 1488-1522)

### 3. Metadata Preservation
- **YouTubePlaylistMenu.kt**: Maps songs to (id, setVideoId) pairs (line 195-205)
- **LocalPlaylistScreen.kt**: Remote sync preserves setVideoId (line 1363-1367)

### 4. All Modified Call Sites (12 total)
- **Database**: DatabaseDao.kt
- **Dialogs**: AddToPlaylistDialog, AddToPlaylistDialogOnline (3 locations), ImportPlaylistDialog
- **Services**: MusicService.kt
- **Menus**: YouTubePlaylistMenu.kt
- **Screens**: LocalPlaylistScreen, OnlinePlaylistScreen, HomeScreen
- **Utils**: SyncUtils.kt
- **ViewModels**: WrappedManager, StatsViewModel

## Testing
- All changes compile successfully (pre-existing MessageCodec.proto errors unrelated)
- Implementation follows Kotlin coroutine best practices
- Atomic transaction blocks prevent partial state corruption
- setVideoId preservation ensures YouTube reordering operations work post-sync

## Related Issues
[...]

Issue #1932, opened 2025-11-18
Playlist cant remove song


SarthakRaymondal:
### Checklist

- [x] I am able to reproduce the bug with the [latest debug version](https://github.com/mostafaalagamy/Metrolist/actions).
- [x] I've checked that there is no open or closed issue about this bug.
- [x] This issue contains only one bug.
- [x] The title of this issue accurately describes the bug.

### Steps to reproduce the bug

1.Create a synced yt playlist
2.Add 2 or more songs
3.Go to the playlist and try to remove any one song
4.Click on sync button 
5.Boom!The deleted song comes back

### Expected behavior

It should permanently delete

### Actual behavior

It comes back cuz it actually only deletes it from app,not yt music

### Logs

No log

### Metrolist version

12.7.0

### Android version

10

koj-yh:
I and many others have had the same issue. The only way afaik to clear songs is from an official ytm/yt client while logged into your account. That's how i delete and shuffle around songs in my playlists.

Hikari2w2:
same issue

Jayesh-69:
Hie, its similar to #1873, and I would like to work here, as its my first time can any one tell me some basic guide if needed?

nyxiereal:
This is resolved on current main: Local playlist removal now schedules the matching remote setVideoId operation; #1013 remains the separate Home/online surface gap.

Closing as completed. If it still reproduces on a current nightly, please comment with the affected media ID where applicable and fresh logs so we can reopen it.

PR #4362, opened 2026-09-11
fix(sync): drop YouTube-deleted songs instead of re-appending them

0Crazy-0:
## Problem
Songs deleted from a playlist in YouTube Music reappear at the end of that playlist in Metrolist after auto-sync, instead of disappearing.

## Cause
In `executeSyncPlaylist` (`SyncUtils.kt`), every local song absent from the remote fetch was unconditionally re-appended at the tail (`preservedSongs`, `position = songs.size + index`) after clearing and repopulating from remote. That set is exactly the songs deleted on YouTube, so remote deletions never propagated and the sync never converged (full rewrite on every cycle).

## Solution
- Only preserve local additions never confirmed on YouTube (`setVideoId == null`, e.g. added offline in Metrolist). Songs with a known `setVideoId` that are missing from remote are treated as intentionally deleted and dropped.
- AI-assisted contribution (declared per `development_guide.md`).

## Testing
- `./gradlew :app:assembleFossDebug`: BUILD SUCCESSFUL
- `./gradlew :app:testFossDebugUnitTest --tests "com.metrolist.music.db.*" --tests "com.metrolist.music.utils.PlaylistSyncTest"`: BUILD SUCCESSFUL
- Manual QA on emulator: (A) delete a song in YT Music → sync → song is gone locally (previously reappeared at the tail); (B) add a song offline → sync → song is kept locally.

## Related Issues
- Related to #1932 (reappearing-song symptom)
- Related to #1873 (same symptom, reported independently)
- Related to #4273 (prior attempt with unconditional preservation; this narrows it via `setVideoId` so YT-side deletions still propagate)
- Related to #3436 (original preservation mechanism, downloaded-songs-only)
- Related to #3446 (parallel mechanism using the `inLibrary` flag, which coexists with this one)
- Related to #4097 (same `setVideoId`-as-truth idea at a different layer; no merge conflict: touches `SyncUtils.kt:1770-1777`, this change only inserts one line at 1471)

## Summary by CodeRabbit

- **Bug Fixes**
  - Playlist synchronization now distinguishes genuinely empty remote playlists from temporarily incomplete or unavailable playlist data.
  - Locally retained songs with an associated video ID are no longer automatically re-added when absent from the remote playlist.
  - This prevents valid remote playlist content from being skipped and avoids unintended entries returning to local playlists during synchronization.


0Crazy-0:
Added `isGenuineEmptyPlaylist(page)`, which only treats the empty fetch as trustworthy when `songCountText` explicitly confirms `"0 songs"` (same regex pattern already used elsewhere in the codebase). When `songCountText` is `null`, meaning there's no way to confirm whether it's a genuine empty playlist or an incomplete/failed fetch, the guard stays in place and the sync is skipped, same as before.

Went with the conservative branch here on purpose: an ambiguous empty result staying unsynced for one cycle is a much cheaper mistake than wiping out a whole playlist based on a read we can't actually verify. Added 4 unit tests covering the combinations (confirmed non-zero count, confirmed zero count, missing count, non-empty fetch).

Issue #3159, opened 2026-03-05
Upcoming song doubles during crossfade


Jenkie03:
### Checklist - Required

- [x]  I am able to reproduce the bug with the [latest debug version](https://github.com/MetrolistGroup/Metrolist/actions/workflows/build.yml?query=branch%3Amain).
- [x]  I've checked that there is NO open or closed issue about this bug. (Duplicate issues will be closed immediately without response)
- [x]  This issue contains only ONE bug. PLEASE check pinned issues before opening a new one.
- [x]  The title of this issue accurately describes the bug.
- [x]  The entire issue report is written/translated in English.

### Steps to reproduce the bug

Enable the Crossfade feature in the application settings.
Set the crossfade duration to any value (the bug occurs regardless of whether it is set to a short or long duration).
Play a song and wait for the transition to the next track.
Observe the audio "doubling" at some point in some songs.

### Expected behavior

A smooth and seamless transition between all tracks. (No doubling)

### Actual behavior

When a crossfade transition occurs, the incoming track does not start smoothly. Instead, the audio at the very beginning of the new song repeats or "doubles" (stutters) before continuing.

The stuttering effect is slightly more pronounced and lasts longer in the screen recording probably due to system overhead, but the core issue is the same.

I have confirmed that this issue is directly linked to the "Skip silence" feature.
If "Skip silence" is ENABLED: The audio doubling/stuttering occurs during every crossfade.
If "Skip silence" is DISABLED: The crossfade works perfectly and transitions are smooth.

### Screenshots/Screen recordings

https://github.com/user-attachments/assets/a2da1ed5-7968-4b4e-94c8-1ba662ad4ae8

### Logs

[04_03-10-55-32_044.log](https://github.com/user-attachments/files/25763985/04_03-10-55-32_044.log)

### Metrolist version

13.2.1

### Android version

16

LenKrazy:
I experienced this myself and the fix for it is to disable "**Skip Silence**" AND "**Instantly Skip Silence**", they are both located underneath the History Duration setting which is also below Crossfade. Hope this helps!

Jenkie03:
> I experienced this myself and the fix for it is to disable "**Skip Silence**" AND "**Instantly Skip Silence**", they are both located underneath the History Duration setting which is also below Crossfade. Hope this helps!

This helped, thx

saivijaychandan:
[...]

PR #4389, opened 2026-09-16
fix(playback): don't pause fading track on crossfade buffering stalls

Llucs:
## Problem
With crossfade enabled, the tail of the current track hitches: it briefly pauses and then comes back mixed over the head of the next track, instead of a smooth overlap. Earlier iterations only treated the symptoms, with no audible improvement.

## Cause
- `startCrossfade()` built a whole second ExoPlayer (`createExoPlayer()` with blocking DataStore reads, full queue copy, `prepare()`) exactly at the trigger instant, on the playback thread, then swapped immediately while the next track was still buffering.
- The swap-time `onIsPlayingChanged` listener paused the fading track on every `isPlaying=false`, including buffering stalls of the not-yet-ready next player.
- ExoPlayer's built-in silence skipping plus `Instant Skip Silence` could `seekTo()` mid-fade, doubling/cutting audio (cf. #3159).

## Solution
- Pre-buffer the crossfade target ~12s before the trigger (`crossfadePrepareJob` + `prepareCrossfadePlayer()` with `playWhenReady=false`); the trigger only swaps. Heavy init/buffering no longer hitch the tail.
- Swap only when the prepared player is `STATE_READY`; otherwise abort the fade and let the normal queue advance handle the transition (a plain transition beats a stutter).
- Crossfade path uses cached prefs (`useCachedPrefs`) instead of blocking DataStore reads; shuffle order uses the cached flag.
- Silence skipping (built-in + instant detector) forced off on both players during the fade, user preference restored on the survivor in `cleanupCrossfade()`; instant-skip rechecks `isCrossfading` before every seek.
- Explicit pause during a fade pauses the fading track via `onPlayWhenReadyChanged`; buffering stalls no longer pause it.
- Single file, no DB/schema/string changes: `MusicService.kt`.

## Testing
- `git diff --check` clean; no new imports/APIs.
- Local full build not possible in this container (no JDK/Android SDK for `./gradlew :app:assembleFossDebug`); relying on CI `Build PR` (`assembleGmsDebug`).
- Manual path on device (needs the PR artifact APK): enable Crossfade + Skip silence + Instantly Skip Silence, play a queue through several transitions; also try pausing mid-fade and a slow network (not-ready target must fall back to a plain transition, never a stutter).

## Related Issues
- Related to #3159
- Related to #3210

## Summary by CodeRabbit

* **Bug Fixes**
  * Silence skipping now stops promptly when a crossfade begins, including during suspended playback.
  * During crossfade, the fading track now pauses when the new track is explicitly paused.
  * Silence-skip checks are re-evaluated before seeking and after playback settles, preventing unwanted skipping during crossfades.
  * Playback transitions now respond more reliably when pausing or changing tracks during a crossfade.


Llucs:
Ignore that PR, I broke everything by accident 😭🙏🏽

PR #4405, opened 2026-09-22
Fix: Support cloud-locker uploaded songs in playlists without artist metadata

RaresB20:
### Problem
When syncing playlists containing user-uploaded cloud tracks (.flac / personal uploads), YouTube Music provides no secondary line runs (`flexColumns[1]` is null), leaving `song.artists` empty. In `SyncUtils.executeSyncPlaylist`, the sync engine aborted inserting these tracks into the local database because `it.artists.isNotEmpty()` failed. Consequently, Room dropped them from the relation query, leaving playlists empty and unplayable.

### Solution
1. In `PlaylistPage.kt`, provide a fallback Artist entity (`"Uploaded"`) when `extractArtists` returns an empty list.
2. In `SyncUtils.kt`, check `existingSongIds` to verify songs are physically stored in Room even when playlist-song mapping counts match.

## Summary by CodeRabbit

- **Bug Fixes**
  - Playlist synchronization now detects and adds remotely available songs that are missing locally, ensuring playlists contain the complete song list.
  - Uploaded tracks without artist metadata now display “Uploaded” instead of appearing without an artist.


RaresB20:
Closing: issue was caused by the 'Hide video songs' preference filtering out uploaded tracks (isVideo = true).

